PR #3, opened 2020-04-22
Remove "sync file events" option.

sam-mccall:
Clangd still doesn't listen to these events, and these days it can
dynamically register file watchers if it wants to start.


sam-mccall:
Yeah it wasn't set up yet when I sent this. It should run on new PRs.

Not happy with the format part yet, it should rather check the diff only.

PR #8, opened 2020-04-30
Automate releasing the extension to VSCode Marketplace from the CI.

### LINKED ISSUE #2: Automate publishing releases
@hokein suggested to trigger this on commits that change the version in `package.json`, which makes sense to me.

hokein:
The idea is to make a new release whenever there is a commit that changes the version in package.json.

Fixes #2 

I verified that the setup works (all steps are executed) in a separate repo. 
Looks like I don't have a permission to add a secret token in `vscode-clangd` repo.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -1,0 +1,28 @@
+name: Publish to VSCode marketplace
+on:
+  push:
+    branches: [ master ]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - name: start X server
+      run: |
+        Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
+        echo "::set-env name=DISPLAY:::99"
+    - uses: actions/checkout@v2
+    - name: detect version
+      uses: EndBug/version-check@v1
+      id: check
+      with:
+        diff-search: true # seach the diff content with version
+    - if: steps.check.outputs.changed == 'true'
+      run: npm ci
+    - if: steps.check.outputs.changed == 'true'
+      run: npm test
+    - name: publish
+      if: steps.check.outputs.changed == 'true'
+      # The token will be expired in on 12/31/2020. Regenerate it at
+      # https://llvm-vs-code-extensions.visualstudio.com/_usersSettings/tokens.
+      run: npm run publish -p "${{ secrets.VSCODE_MARKETPLACE_TOKEN }}"
+
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,9 +16,9 @@
             "integrity": "sha512-1Bh06cbWJUHMC97acuD6UMG29nMt0Aqz1vF3guLfG+kHHJhy3AyohZFFxYk2f7Q1SQIrNwvncxAE0N/9s70F2w==",
             "dev": true,
             "requires": {
-                "@types/events": "*",
-                "@types/minimatch": "*",
-                "@types/node": "*"
+                "@types/events": "3.0.0",
+                "@types/minimatch": "3.0.3",
+                "@types/node": "6.14.9"
             }
         },
         "@types/minimatch": {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
       with:
         source: 'src test'
         extensions: 'ts'
+
+  publish:
+    name: Publish to VSCode marktplace
```
sam-mccall, line 26: nit: marketplace

.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: contains(github.ref, 'master') # publish it only if the push comes from master.
```
sam-mccall, line 28: I think it'd be nice to isolate this to a different workflow with different triggers, even if it means ignoring the CI.

If we want to run the tests, maybe we should do that explicitly (and redundantly) in the separate workflow, it's not much code.
I'm not sure it makes any sense to block releases on clang-format.

.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: detect version
+      uses: EndBug/version-check@v1
+      id: check
+      with:
+        diff-search: true # seach the diff content with version
+        file-name: package.json
+        token: ${{ secrets.GITHUB_TOKEN }}
```
sam-mccall, line 39: you don't need this for a public repository, per the docs

.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
...
+    needs: build # make sure the build is passed.
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: detect version
+      uses: EndBug/version-check@v1
+      id: check
+      with:
+        diff-search: true # seach the diff content with version
+        file-name: package.json
```
sam-mccall, line 38: this is the default value

.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
...
+        token: ${{ secrets.GITHUB_TOKEN }}
+    - name: version bumped
```
sam-mccall, line 40: I don't think this step is so useful, the following steps run on the same condition, and it's easy to distinguish steps that are run vs skipped in the github UI

.github/workflows/ci.yml
hokein: oh, right, I was testing it with a private repo, without this token, it throwed some errors.

.github/workflows/auto-release.yml
```diff
@@ -0,0 +1,26 @@
...
+      id: check
+      with:
+        diff-search: true # seach the diff content with version
+    - if: steps.check.outputs.changed == 'true'
+      run: npm ci
+    - if: steps.check.outputs.changed == 'true'
+      run: npm test
+    - name: publish
+      if: steps.check.outputs.changed == 'true'
+      run: npm run publish -p ${{ secrets.VSCODE_MARKETPLACE_TOKEN }}
```
sam-mccall, line 25: nit: quote the token

hokein:
I think it should work, but I didn't use it to make a real release (shall we bump a version for testing?)

Just a note: the secret is added, but it will be expired in on 12/31/2020. Regenerate it at https://llvm-vs-code-extensions.visualstudio.com/_usersSettings/tokens when expired.

sam-mccall:
A real release would be good actually - e.g. for 0e41bf23184f213b317ccd03fe28af9efd2c95c1

So merge this and try to make a release?

> Just a note: the secret is added, but it will be expired in on 12/31/2020. Regenerate it at https://llvm-vs-code-extensions.visualstudio.com/_usersSettings/tokens when expired.

Can you add this as a comment to the yaml workflow?

(BTW this is going to bite us the first time we try to push a release after it expires: the workflow will fail and we don't have a way to force it to rerun other than bumping the version again. Which is probably a fine workaround)

hokein:
> A real release would be good actually - e.g. for [0e41bf2](https://github.com/clangd/vscode-clangd/commit/0e41bf23184f213b317ccd03fe28af9efd2c95c1)
> 
> So merge this and try to make a release?

sounds good.

> Can you add this as a comment to the yaml workflow?

Done.

> (BTW this is going to bite us the first time we try to push a release after it expires: the workflow will fail and we don't have a way to force it to rerun other than bumping the version again. Which is probably a fine workaround)

yeah, we could do `git push -f` :), or setup an alarm when the token is going to expired.

PR #9, opened 2020-04-30
Make sure the code is clang-format clean.

hokein:


src/extension.ts
```diff
@@ -58,20 +58,19 @@ export function activate(context: vscode.ExtensionContext) {
...
-      { scheme: 'file', language: 'cpp' },
+    documentSelector : [
+      {scheme : 'file', language : 'c'}, {scheme : 'file', language : 'cpp'},
       // CUDA is not supported by vscode, but our extension does supports it.
-      { scheme: 'file', language: 'cuda' },
-      { scheme: 'file', language: 'objective-c' },
-      { scheme: 'file', language: 'objective-cpp' }
+      {scheme : 'file', language : 'cuda'},
+      {scheme : 'file', language : 'objective-c'},
+      {scheme : 'file', language : 'objective-cpp'}
```
sam-mccall, line 66: we should use a trailing comma to get one-per-line instead

src/extension.ts
```diff
@@ -58,20 +58,19 @@ export function activate(context: vscode.ExtensionContext) {
 
   const clientOptions: vscodelc.LanguageClientOptions = {
     // Register the server for c-family and cuda files.
-    documentSelector: [
-      { scheme: 'file', language: 'c' },
-      { scheme: 'file', language: 'cpp' },
+    documentSelector : [
```
sam-mccall, line 61: so this formatting with space before the : is just wrong - it's not what clang-format is trying to do, it's just formatting [ ... ] as type-annotations because we've triggered a bug. Let me have a look at whether there's a good workaround...

sam-mccall: Actually this is a style option - it's just a terrible one.
See #11 - WDYT?

hokein:
Closing it in favor of #11

PR #20, opened 2020-05-11
Type hierarchy support

HighCommander4:
This is a first draft which I'm posting for feedback.

Currently, the type hierarchy view only shows derived classes (`Children` direction), which is the more useful one (since base classes can be navigated to from the base clauses in the class definition).

We will need to make a UX choice around how to incorporate the `Parents` direction as well. My suggestion is:

  * Stick to just one view and one command to activate it.
  * The view shows one of the directions at a time, and Indicates which one it's currently showing.
  * There is a button to cause the view to switch to showing the other direction.

This is how vscode's built-in call hierarchy view works (for showing callers vs. callees), and given the similarity in the protocols, I am venturing a guess that this is how vscode's built-in type hierarchy view (which doesn't exist yet, but will eventually replace this when it does) will work as well.

Thoughts / feedback on this direction is welcome.


sam-mccall:
Oh, this is neat!

Random observations from playing with it:
 - having it display `test.c:3` is much less useful than having it jump there (I think single-click is fine, like VSCode's call hierarchy). Once this happens we can probably drop the filename.
 - mirroring call hierarchy for the directions seems like a reasonable call, though I didn't find that UI particularly intuitive. It's really tempting to include nodes for the parents (in the case where there's exactly one parent)
 - the "reveal" isn't working for me: it focuses the left panel, but doesn't expand or focus the type hierarchy section.
 - the context menu item should probably be in the references section if possible

I haven't looked deeply at the code but it looks reasonable.
This really makes me want to export "show AST" as a real feature...

HighCommander4:
Thanks for the feedback! I pushed an update that addresses many of the comments.

>    * having it display `test.c:3` is much less useful than having it jump there (I think single-click is fine, like VSCode's call hierarchy). Once this happens we can probably drop the filename.

I implemented single-click navigation.

The filename and line number are still there. I'm somewhat ambivalent about removing them, as they can still be useful (e.g. you could have different derived classes by the same name in different files). Perhaps keeping the filename only would make sense?

>    * mirroring call hierarchy for the directions seems like a reasonable call

Haven't implemented this yet, will do in the next update.

> It's really tempting to include nodes for the parents (in the case where there's exactly one parent)

I agree it's really tempting. But what do you do when there are multiple parents? If you show just one, or none, we might be misleading the user.

>    * the "reveal" isn't working for me: it focuses the left panel, but doesn't expand or focus the type hierarchy section.

Yeah, the logic in the first draft only caused the view to be shown (i.e. have it appear in the panel), but not expanded or focused.

I tried harder in this draft, should expand + focus now.

>    * the context menu item should probably be in the references section if possible

Fixed.

> This really makes me want to export "show AST" as a real feature...

Just as another consumer of the tree view facility? Or is there some other connection?

package.json
```diff
@@ -122,6 +130,31 @@
...
                 "when": "editorTextFocus"
             }
-        ]
+        ],
+        "menus": {
+            "editor/context": [
+                {
+                    "command": "clangd.typeHierarchy",
+                    "when": "resourceLangId == cpp",
+                    "group": "0_navigation@4"
```
HighCommander4, line 139: What's this, you ask? This is how I got it to be placed in the [references section](https://github.com/microsoft/vscode-references-view/blob/f63eaed9934ca5ecc8f3fb3ca096f38c6e5e181f/package.json#L162).

I'd write this in a comment, but... JSON doesn't support comments.

HighCommander4:
> > * mirroring call hierarchy for the directions seems like a reasonable call
> 
> Haven't implemented this yet, will do in the next update.

This is done now as well. I think the patch should be ready for review.

package.json
```diff
@@ -122,6 +150,51 @@
...
+                {
+                  "command": "clangd.typeHierarchy.close",
+                  "when": "view == clangd.typeHierarchyView",
+                  "group": "navigation"
+                }
+            ],
+            "commandPalette": [
+                {
+                    "command": "clangd.typeHierarchy.viewParents",
+                    "when": "false"
```
sam-mccall, line 182: FWIW call hierarchy exposes the equivalents in the command palette.

It calls these "Calls: Show call hierarchy"/"Calls: Show incoming calls" etc.
We could do something similar with a "Types" prefix...

package.json
sam-mccall: you might get away with adding a "_comment" key, but up to you whether that's too horrible

package.json
```diff
@@ -113,6 +113,34 @@
...
+            {
+                "command": "clangd.typeHierarchy.close",
+                "title": "Close",
+                "icon": {
+                  "light": "resources/close-light.svg",
+                  "dark": "resources/close-dark.svg"
+                }
             }
         ],
         "keybindings": [
```
sam-mccall, line 146: You may want to bind a key for this, call hierarchy is shift-alt-h. Is shift-alt-t free?

resources/derived-dark.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path fill="#E8E8E8" d="M5.1 5l-.9.9 3.6 3.6 3.9-3.6-1-.9-3 2.7L5.1 5z"/></svg>
```
sam-mccall, line 1: I'd be tempted to call these down/up so we can reuse them...

src/type-hierarchy.ts
```diff
@@ -0,0 +1,218 @@
...
+        'typeHierarchy/resolve');
+}
+
+class TypeHierarchyTreeItem extends vscode.TreeItem {
+  constructor(item: TypeHierarchyItem) {
+    super(item.name);
+    if (item.children && item.children.length > 0) {
+      this.collapsibleState = vscode.TreeItemCollapsibleState.Expanded;
+    } else {
+      this.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
```
sam-mccall, line 57: don't show the disclosure widget if we know there are no children.
i.e. children != null && children.length == 0 ==> TreeItemCollapsibleState.None

src/type-hierarchy.ts
```diff
@@ -0,0 +1,218 @@
...
+    const filepath = vscode.Uri.parse(item.uri).path;
+    this.description =
+        `${path.basename(filepath)}:${item.range.start.line + 1}`;
+
+    // Make the item respond to a single-click by navigating to the
+    // definition of the class.
+    this.command = {
+      arguments : [ item ],
+      command : 'clangd.typeHierarchy.gotoItem',
+      title : 'Goto'
```
sam-mccall, line 70: nit: Go to, two words

src/type-hierarchy.ts
```diff
@@ -0,0 +1,218 @@
...
+    editor.selection = new vscode.Selection(range.start, range.end);
+  }
+
+  public async setDirection(direction: TypeHierarchyDirection) {
+    this.direction = direction;
+    this._onDidChangeTreeData.fire();
+  }
+
+  public getTreeItem(element: TypeHierarchyItem): vscode.TreeItem
+      |Thenable<vscode.TreeItem> {
```
sam-mccall, line 135: no need for the thenable case?

src/type-hierarchy.ts
```diff
@@ -0,0 +1,218 @@
...
+            direction : TypeHierarchyDirection.Children,
+            resolve : 1
+          });
+      element.children = resolved.children;
+    }
+    return element.children;
+  }
+
+  private async reveal(editor: vscode.TextEditor) {
+    console.log("reveal()");
```
sam-mccall, line 171: drop this :-)

src/type-hierarchy.ts
```diff
@@ -0,0 +1,218 @@
...
+      // Resolve both directions initially. That way, if the user switches
+      // to the Parents view, we have the data already. Note that clangd
+      // does not support resolving parents via typeHierarchy/resolve,
+      // so otherwise we'd have to remember the TextDocumentPositionParams
+      // to make another textDocument/typeHierarchy request when switching
+      // to Parents view.
+      direction : TypeHierarchyDirection.Both
+    });
+    if (item) {
+      console.log("got item with " + item.children.length + " children");
```
sam-mccall, line 194: remove

sam-mccall:
> The filename and line number are still there. I'm somewhat ambivalent about removing them, as they can still be useful

The filename will be truncated most of the time: the panel is by default quite narrow, with indentation followed by the arrow indicator, followed by the name of the class, and finally the filename. If either the classname or the filename are long at all, it's going to be truncated. Once the extension is truncated it loses a lot of value as you can't tell it's a filename.

I really think this is just going to be visual noise. It's also inconsistent with other places where we show classnames (like the workspace symbols panel) where the small grey text is the namespace rather than the filename.

> I agree it's really tempting. But what do you do when there are multiple parents? If you show just one, or none, we might be misleading the user.

I think we could show a dummy node "multiple parents" - it's not really elegant, but this is a rare case and it'd be great to avoid the awkward "reverse direction" interaction in the common case.

HighCommander4:
Addressed review comments.

HighCommander4:
> > I agree it's really tempting. But what do you do when there are multiple parents? If you show just one, or none, we might be misleading the user.
> 
> I think we could show a dummy node "multiple parents" - it's not really elegant, but this is a rare case and it'd be great to avoid the awkward "reverse direction" interaction in the common case.

I implemented this in a separate commit: https://github.com/HighCommander4/vscode-clangd/pull/1/commits

HighCommander4:
> I implemented this in a separate PR: https://github.com/HighCommander4/vscode-clangd/pull/1/commits

Er, whoops, didn't mean to create a PR against my own fork... let's try that again: https://github.com/clangd/vscode-clangd/pull/44

HighCommander4:
> > I implemented this in a separate PR: https://github.com/HighCommander4/vscode-clangd/pull/1/commits
> 
> Er, whoops, didn't mean to create a PR against my own fork... let's try that again: #44

Not quite wanted I wanted either, as that PR contains both commits.

Sorry, the GitHub isn't strong with me: what's the proper way to create a PR against this main fork which depends on another PR (i.e. this one), and only contains that second commit?

Trass3r:
Not possible. https://github.com/isaacs/github/issues/959#issuecomment-642834187

HighCommander4:
Review ping :)

hokein:
sorry, I planned to do the review, but somehow got distracted and forgot it. Will try to review it today or tomorrow.

> Sorry, the GitHub isn't strong with me: what's the proper way to create a PR against this main fork which depends on another PR (i.e. this one), and only contains that second commit?

I think you could create two branches in "clangd/vscode-clangd" repo (you should have permission), and send a PR against two branches.

HighCommander4:
> I think you could create two branches in "clangd/vscode-clangd" repo (you should have permission), and send a PR against two branches.

Thanks, will try that next time. For now, if you'd like to see the parts of the patch that have been split out to implement the "try to show parents as well" suggestion, you can look at just the second commit in #44.

src/extension.ts
```diff
@@ -132,6 +133,7 @@ export async function activate(context: vscode.ExtensionContext) {
   if (config.get<boolean>('semanticHighlighting'))
     semanticHighlighting.activate(client, context);
   client.registerFeature(new EnableEditsNearCursorFeature);
+  new typeHierarchy.TypeHierarchyProvider(context, client);
```
hokein, line 136: We use the "activate" pattern for new features, could you follow it (see `semanticHighlighting.activate` for reference)?

src/type-hierarchy.ts
```diff
@@ -0,0 +1,213 @@
...
+  deprecated?: boolean;
+  uri: string;
+  range: vscodelc.Range;
+  selectionRange: vscodelc.Range;
+  parents?: TypeHierarchyItem[];
+  children?: TypeHierarchyItem[];
+  data?: any;
+}
+
+export namespace TypeHierarchyRequest {
```
hokein, line 31: this doesn't need to be exported? it is just used in this file. the same to `ResolveTypeHierarchyRequest`

src/type-hierarchy.ts
```diff
@@ -0,0 +1,213 @@
+import * as path from 'path';
```
hokein, line 1: given that type-hierarchy is not in LSP yet, would be nice to have a high-level comment on the top of this file, e.g.  give a reference link of the type-hierarchy proposal.

src/type-hierarchy.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  private close() {
+    // Hide the type hierarchy view.
+    vscode.commands.executeCommand(
+        'setContext', 'extension.vscode-clangd.typeHierarchyVisible', false);
+
+    this.root = undefined;
+    this._onDidChangeTreeData.fire();
+  }
+}
```
hokein, line 213: nit: add a newline at end of the file.

src/type-hierarchy.ts
```diff
@@ -0,0 +1,213 @@
...
+import * as vscode from 'vscode';
+import * as vscodelc from 'vscode-languageclient';
+
+export namespace TypeHierarchyDirection {
+export const Children = 0;
+export const Parents = 1;
+export const Both = 2;
+}
+
+type TypeHierarchyDirection = 0|1|2;
```
hokein, line 11: `TypeHierarchyDirection` is used for namespace above, can we use a different name for type?

HighCommander4: My understanding was that this pairing of a namespace and a type with the same name is a pattern in the vscode API.

This is how `TypeHierarchyDirection` is defined in the [type hierarchy protocol proposal](https://github.com/microsoft/vscode-languageserver-node/pull/426/commits/876d7fe224e17d7c08258a6da21d11a2df9c1d0d#diff-0b22db02bb4adb043694d9d0a11d0aefR83), as the pattern is used for other enum-like things in the existing API (e.g. [`TextDocumentSyncKind`](https://github.com/microsoft/vscode-languageserver-node/blob/bf091f5bff33797888aa8b14b09cae918e88f589/protocol/src/common/protocol.ts#L1036)).

HighCommander4:
Thanks for the review, I believe I addressed all of the comments.

> btw, I saw an exception in the debug console when playing around this patch locally, no sure which operation caused it.

Based on the stack trace, it is probably caused by the call to `this.treeView.reveal()` in `TypeHierarchyProvider.reveal()`. However, I have not been able to trigger it.

It's worth noting that #44 makes changes that could fix this (because we now have to support `reveal()` for any node, not just the root, and accordingly it implements `getParent()` for all nodes).

src/type-hierarchy.ts
```diff
@@ -0,0 +1,224 @@
...
+    // definition of the class.
+    this.command = {
+      arguments: [item],
+      command: 'clangd.typeHierarchy.gotoItem',
+      title: 'Go to'
+    };
+  }
+}
+
+export class TypeHierarchyProvider implements
```
hokein, line 85: nit: no need to export it.

HighCommander4:
> btw, please don't force-push commits, it makes reviewers harder to see the difference, I'd just append new commits.

Sorry, still getting used to the GitHub workflow which seems quite different from the Phabricator one (where I do `--amend` after each set of comments and resubmit).

I assume I _should_ squash the commits together before merging, though, to avoid polluting the repo history with "Address review comments" commits.

Issue #22, opened 2020-05-11
Support relative clangd path


apockoffork:
My project (which has 100s of easily confused users) has a configured value for `clangd.path`. When I start up vscode now it prompts asking if I want to install clangd. It would be great if the extension could  skip prompting if the path is valid or provide a config option to disable the prompt.

hokein:
just set `"clangd.checkUpdates": false` in the setting.json to disable it.

sam-mccall:
Hmm, no, this sounds more like the autoinstall flow than the auto update flow.

@apockoffork the extension does check if the configured path exists, can you give a little more detail? (Is the configured path the default "clangd" to lookup on PATH, or an absolute path? Does this affect a particular platform? Etc)

apockoffork:
Thanks for the super fast reply. Upon further investigation it seems there is more than one thing probably going on here. 

The `checkUpdates` flag is complaining about the config type
![image](https://user-images.githubusercontent.com/10534342/81582093-9d735d80-937d-11ea-86aa-81925355dd44.png)
That complaint continues on the settings json:
![image](https://user-images.githubusercontent.com/10534342/81582481-1c689600-937e-11ea-80f2-ff5d449408ff.png)

We use a relative path to our clangd executable:
`"clangd.path": "tools/clangd",` 
That wasn't an issue previously but perhaps things have changed and now thats an issue.
The other quirk of our setup is that `tools/clangd` is actually wrapper around the actual clangd executable (it runs like clangd but it isn't technically the clangd binary) so if you're doing some sort of inspection of the file to determine what version of clangd it is or something like that it will probably fail for us.

sam-mccall:
> The checkUpdates flag is complaining about the config type

Oh dear, the default is `"false"` instead of `false`, which means... true :-(
I've pushed 0.1.3 to fix this. Thanks!

> We use a relative path to our clangd executable

Ah, and this is relative to `$PATH`, rather than the working directory?
I'm not surprised this doesn't work after the patch (but a little surprised that it worked before). 
I don't think this is standard, e.g. `bash` doesn't seem to accept this:

```
sammccall@pope:~$ bin/xx/foo.sh
foosh
sammccall@pope:~$ PATH=$PWD/bin/xx
sammccall@pope:~$ foo.sh
foosh
sammccall@pope:~$ PATH=$PWD/bin
sammccall@pope:~$ xx/foo.sh
bash: xx/foo.sh: No such file or directory
```

[...]

PR #24, opened 2020-05-11
Support a subset of substitutions like ${workspaceRoot} in config.

sam-mccall:
VSCode doesn't currently provide any APIs for extensions to expand these.
So we pick a few that seem useful and are easy to substitute.

Fixes https://github.com/clangd/vscode-clangd/issues/22

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -18,7 +18,7 @@ function substitute<T>(val: T): T {
   if (typeof val == 'string') {
     val = val.replace(/\$\{(.*?)\}/g, (match, name) => {
       const rep = replacement(name);
-      // If there's no replacement,available, keep the placeholder.
+      // If there's no replacement available, keep the placeholder.
       return (rep === null) ? match : rep;
     }) as unknown as T;
   } else if (Array.isArray(val))
@@ -45,12 +45,13 @@ function replacement(name: string): string|null {
   }
   if (name == 'cwd')
     return process.cwd();
-  if (name == 'execPath')
-    return vscode.env.appRoot;
-  if (name.startsWith('env:'))
-    return process.env[name.substr(4)] || '';
-  if (name.startsWith('config:')) {
-    const config = vscode.workspace.getConfiguration().get(name.substr(7));
+  const envPrefix = 'env:';
+  if (name.startsWith(envPrefix))
+    return process.env[name.substr(envPrefix.length)] || '';
+  const configPrefix = 'config:';
+  if (name.startsWith(configPrefix)) {
+    const config = vscode.workspace.getConfiguration().get(
+        name.substr(configPrefix.length));
     return (typeof config == 'string') ? config : null;
   }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -36,15 +36,13 @@ function substitute<T>(val: T): T {
 // Subset of substitution variables that are most likely to be useful.
 // https://code.visualstudio.com/docs/editor/variables-reference
 function replacement(name: string): string|null {
-  if (name == 'workspaceRoot' || name == 'workspaceFolder') {
+  if (name == 'workspaceRoot' || name == 'workspaceFolder' || name == 'cwd') {
     if (vscode.workspace.rootPath !== undefined)
       return vscode.workspace.rootPath;
     if (vscode.window.activeTextEditor !== undefined)
       return path.dirname(vscode.window.activeTextEditor.document.uri.fsPath);
     return process.cwd();
   }
-  if (name == 'cwd')
-    return process.cwd();
   const envPrefix = 'env:';
   if (name.startsWith(envPrefix))
     return process.env[name.substr(envPrefix.length)] || '';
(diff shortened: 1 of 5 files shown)


src/config.ts
```diff
@@ -0,0 +1,58 @@
...
+function replacement(name: string): string|null {
+  if (name == 'workspaceRoot' || name == 'workspaceFolder') {
+    if (vscode.workspace.rootPath !== undefined)
+      return vscode.workspace.rootPath;
+    if (vscode.window.activeTextEditor !== undefined)
+      return path.dirname(vscode.window.activeTextEditor.document.uri.fsPath);
+    return process.cwd();
+  }
+  if (name == 'cwd')
+    return process.cwd();
```
hokein, line 47: so if I open `vscode-clangd` project in vscode, `${cwd}` should be `/path/to/vscode-clangd`, but `process.cwd()` returns a `/` on my local machine :(

src/config.ts
```diff
@@ -0,0 +1,58 @@
...
+                          target?: vscode.ConfigurationTarget) {
+  return vscode.workspace.getConfiguration('clangd').update(key, value, target);
+}
+
+// Traverse a JSON value, replacing placeholders in all strings.
+function substitute<T>(val: T): T {
+  if (typeof val == 'string') {
+    val = val.replace(/\$\{(.*?)\}/g, (match, name) => {
+      const rep = replacement(name);
+      // If there's no replacement,available, keep the placeholder.
```
hokein, line 21: nit: there is a typo between replacement and available, should be ` `.

src/extension.ts
```diff
@@ -53,9 +44,9 @@ export async function activate(context: vscode.ExtensionContext) {
 
   const clangd: vscodelc.Executable = {
     command: clangdPath,
```
hokein, line 46: by reading the LSP client [code](https://github.com/microsoft/vscode-languageserver-node/blob/635ab1fe6f8c57ce9402e573d007f24d6d290fd3/client/src/main.ts#L468), it seems that it will try to find a working directory (default is the workspacefolder) as the command `cwd`, that might explain why the relative path was working prior to the auto-install patch.

src/config.ts
```diff
@@ -0,0 +1,58 @@
...
+    val = result as T;
+  }
+  return val;
+}
+
+// Subset of substitution variables that are most likely to be useful.
+// https://code.visualstudio.com/docs/editor/variables-reference
+function replacement(name: string): string|null {
+  if (name == 'workspaceRoot' || name == 'workspaceFolder') {
+    if (vscode.workspace.rootPath !== undefined)
```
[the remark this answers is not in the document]
sam-mccall: I'm not really sure what you want me to do here: we can't substitute an array here.
${workspaceRoot} is the substitution that was specifically requested, and it corresponds to vscode.workpace.rootPath.

That does match the first entry of workspaceFolder if it exists, but I don't think writing it that way would be clearer, would it?

IOW, it's the *concept* of workspaceRoot that's deprecated, but I don't see how we can avoid it here.

src/config.ts
```diff
@@ -0,0 +1,58 @@
+import * as path from 'path';
+import * as process from 'process';
```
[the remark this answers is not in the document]
sam-mccall: The import is required for TS, not node. (There's no `require` here)

src/extension.ts
sam-mccall: Yeah, I understand why it happened to work before, but I don't think we should continue to support it - it would be a bit of work (since we access it in a few places now) and seems pretty fragile.

src/config.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (vscode.workspace.rootPath !== undefined)
+      return vscode.workspace.rootPath;
+    if (vscode.window.activeTextEditor !== undefined)
+      return path.dirname(vscode.window.activeTextEditor.document.uri.fsPath);
+    return process.cwd();
+  }
+  if (name == 'cwd')
+    return process.cwd();
+  if (name == 'execPath')
+    return vscode.env.appRoot;
```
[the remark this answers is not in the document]
sam-mccall: Just removed execPath, nobody asked for it and it seems complicated.

src/config.ts
sam-mccall: I can't reproduce this. Are you opening `vscode` from the terminal? If you're opening it from a GUI (e.g. Code.app in Finder) I wouldn't expect the working directory to be meaningful

src/config.ts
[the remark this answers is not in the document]
sam-mccall: Huh, you're right. Removed.

src/config.ts
sam-mccall: Indeed, thanks for pointing this out.
I've made it do the same as workspaceRoot which seems to be basically right except for the error cases.

src/extension.ts
sam-mccall: Now I understood what you meant by "(default is the workspacefolder)", maybe this is actually worth preserving. In a followup patch...

hanct:
So with this changes, now extension developer can use finally use variable expansions on the configuration points that they added to the workspace settings file?

PR #26, opened 2020-05-12
Interpret clangd.path relative to ${cwd} if it has a slash

### LINKED ISSUE #22: Support relative clangd path

sam-mccall:
This doesn't make a lot of sense, but seems to be backwards-compat behavior.

Fixes #22

### DIFF
--- a/src/install.ts
+++ b/src/install.ts
@@ -107,8 +107,14 @@ class UI {
   }
 
   async promptInstall(version: string) {
-    const message = 'The clangd language server was not found on your PATH.\n' +
-                    `Would you like to download and install clangd ${version}?`;
+    const p = this.clangdPath;
+    let message = '';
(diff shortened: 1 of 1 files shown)


src/install.ts
```diff
@@ -111,7 +113,15 @@ class UI {
       common.installLatest(this);
   }
 
-  get clangdPath(): string { return config.get<string>('path'); }
+  get clangdPath(): string {
+    const p = config.get<string>('path');
```
hokein, line 117: nit: const => let, a const can not be changed.

sam-mccall: Done, and actually tested this time...

PR #32, opened 2020-05-18
Make workspaceSymbols work on qualified name.

### LINKED ISSUE #31: Support qualified names in workspace/symbol
Currently, a query like `Class::Method` does not return any results, but it would be useful if it did. The alternatives of searching for `Class` and then manually navigating to `Method` within the class, or searching for `Method` and choosing among potentially many classes that declare a method with that name, are more labour-intensive.

hokein:
Fixes #31 

Tested on LLVM project, verified with qualified names and unqualified names.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -96,7 +96,20 @@ export async function activate(context: vscode.ExtensionContext) {
               return item;
             })
             return new vscode.CompletionList(items, /*isIncomplete=*/ true);
-          }
+          },
+      // VSCode applies fuzzy match only on the symbol name, thus it throws away
+      // all results if query token is a prefix qualified name.
+      // By adding the containerName to the symbol name, it prevents VSCode from
+      // filtering out any results, e.g. enable workspaceSymbols for qualified
+      // symbols.
+      provideWorkspaceSymbols: async (query, token, next) => {
+        let symbols = await next(query, token);
+        return symbols.map(symbol => {
+          if (symbol.containerName)
+            symbol.name = `${symbol.containerName}::${symbol.name}`;
+          return symbol;
+        })
+      },
     },
   };
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -107,6 +107,8 @@ export async function activate(context: vscode.ExtensionContext) {
         return symbols.map(symbol => {
           if (symbol.containerName)
             symbol.name = `${symbol.containerName}::${symbol.name}`;
+          // Always clean the containerName to avoid displaying it twice.
+          symbol.containerName = '';
           return symbol;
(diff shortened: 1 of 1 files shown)


kadircet:
just a drive by comment. i remember this was working in vscode last time i checked (which was a looong time ago). so are we sure this is not a regression on their side (not saying we shouldn't fix it in our extension, but maybe we should also file a bug to them) ?

hokein:
> just a drive by comment. i remember this was working in vscode last time i checked (which was a looong time ago). so are we sure this is not a regression on their side (not saying we shouldn't fix it in our extension, but maybe we should also file a bug to them) ?

Not sure this is a regression, personally I didn't often use this feature in VSCode as it didn't give quite useful results (maybe because of this bug?).

The closest issue I found is https://github.com/microsoft/vscode/issues/23509, it looks like they suggest to address it in the extension side.

src/extension.ts
```diff
@@ -96,7 +96,20 @@ export async function activate(context: vscode.ExtensionContext) {
...
+          },
+      // VSCode applies fuzzy match only on the symbol name, thus it throws away
+      // all results if query token is a prefix qualified name.
+      // By adding the containerName to the symbol name, it prevents VSCode from
+      // filtering out any results, e.g. enable workspaceSymbols for qualified
+      // symbols.
+      provideWorkspaceSymbols: async (query, token, next) => {
+        let symbols = await next(query, token);
+        return symbols.map(symbol => {
+          if (symbol.containerName)
```
sam-mccall, line 108: This results in containerName being displayed twice (once as part of Name and once separately).

image

I'd suggest:

 - only do this if there's a container name *and* the query contains ::
 - clear out containerName when you do this
 - leave a comment explaining that we don't want to display containerName twice

This isn't so bad as the aligned blue in the LHS will still make it easy to visually scan the rest of the symbol names.

sam-mccall:
I filed https://github.com/microsoft/vscode/issues/98125 asking for a real fix in VSCode, but this workaround seems fine, we should do it only in the cases necessary.

src/extension.ts
hokein: > This results in containerName being displayed twice (once as part of Name and once separately).

this is a good point. It is not good to displayed the container name twice. I find the `symbol.containerName` being displayed in a less discoverable way (compare with the `symbol.Name`), what do you think that we always clean the container name, and use the `qualified name` as the `SymbolName`

sam-mccall: Yeah actually I think you're right, this looks better, since the name gets highlighted anyway and now the namespace doesn't run together into the filename.

PR #35, opened 2020-05-27
Add "clangd.restart" command.

hokein:
Similar to YCM's "RestartServer" command.

If clangd crashes, vscode will try to restart clangd up to 5 times. After that, there is no way to restart clangd except reloading the vscode window. However, reloading the vscode window will lose the whole working context in vscode, especially you are running programs in the integrated terminal.

"clangd.restart" command will restart clangd server **without** reloading the vscode window.

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,10 @@
             {
                 "command": "clangd.activate",
                 "title": "Manually activate clangd extension"
+            },
+            {
+                "command": "clangd.restart",
+                "title": "Restart the clangd language server"
             }
         ],
         "keybindings": [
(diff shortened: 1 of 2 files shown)


benjaminhdavis:
We may need the dispose call similar to the vscode-go plugin's restart found here https://github.com/microsoft/vscode-go/pull/2530/files

hokein:
> We may need the dispose call similar to the vscode-go plugin's restart found here https://github.com/microsoft/vscode-go/pull/2530/files

Thanks, this is a good point.

PR #54, opened 2020-07-09
support proposed official semantic highlighting API

Trass3r:
For the record, this seems to be enough to enable support for the new semantic highlighting.
More of an RFC. It needs to be checked what [changed between LSP client 6.1.3 and the final version](https://github.com/microsoft/vscode-languageserver-node/commits/master/protocol/src/common/protocol.semanticTokens.proposed.ts) and which state will be included in the clangd-11 release (v10 falls back to the old implementation).
Even works in non-insiders vscode @sam-mccall.

It's also unclear if
```diff
-        "vscode": "^1.41.0"
+        "vscode": "^1.44.0"

-        "@types/vscode": "1.41.*",
+        "@types/vscode": "1.44.*",
```
offers anything due to https://code.visualstudio.com/updates/v1_44#_semantic-tokens-provider-api

### DIFF
--- a/src/semantic-highlighting.ts
+++ b/src/semantic-highlighting.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs';
 import * as jsonc from 'jsonc-parser';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import * as vscodelc from 'vscode-languageclient';
+import * as vscodelc from 'vscode-languageclient/node';
 import * as vscodelct from 'vscode-languageserver-types';
 
 export function activate(client: vscodelc.LanguageClient,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1555,25 +1555,25 @@
             }
         },
         "vscode-jsonrpc": {
-            "version": "6.0.0-next.4",
-            "resolved": "https://registry.npmjs.org/vscode-jsonrpc/-/vscode-jsonrpc-6.0.0-next.4.tgz",
-            "integrity": "sha512-nIXUeSMKfqvqSG75s1JuUy0UgjE3d0NERjGz7AUGDFXX0/roVUEbaio5VBdus30nty8v0FCf/6MbCSrH5RybcQ=="
+            "version": "6.0.0-next.5",
+            "resolved": "https://registry.npmjs.org/vscode-jsonrpc/-/vscode-jsonrpc-6.0.0-next.5.tgz",
+            "integrity": "sha512-IAgsltQPwg/pXOPsdXgbUTCaO9VSKZwirZN5SGtkdYQ/R3VjeC4v00WTVvoNayWMZpoC3O9u0ogqmsKzKhVasQ=="
         },
         "vscode-languageclient": {
-            "version": "7.0.0-next.8",
-            "resolved": "https://registry.npmjs.org/vscode-languageclient/-/vscode-languageclient-7.0.0-next.8.tgz",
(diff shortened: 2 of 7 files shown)


sam-mccall:
Thanks for linking to all the context! Really useful.

AIUI client v6.13 + "enable proposed features" works with clangd at head (that's what I tested with originally), but the client master uses a silghtly different protocol (semanticTokens/full/delta rather than semanticTokens/edits) and that's the one that will be standardized (described LSP 3.16 draft).
And they use mostly the same capability names so there can be confusion.
Also, we're less than a week away from an LLVM release branch cut, so we need to decide on clangd's behaviour...

I think it's best to implement the likely-final protocol from the draft, and soon (before the release cut):  https://reviews.llvm.org/D83556.
If we do that, this patch won't work with clangd at head anymore (it'll try to speak the wrong protocol), so we again need to wait for a client release that supports it (7.0.0-next.6 was too early and 7.0.0-next.5.1 is on a branch that doesn't have https://github.com/microsoft/vscode-languageserver-node/commit/66f91b3583a19ae621214d908ce19c9b2d182626)

Does that make sense?

Trass3r:
> I think it's best to implement the likely-final protocol from the draft, and soon (before the release cut): https://reviews.llvm.org/D83556.

Yes definitely.

Trass3r:
@sam-mccall It's available now.

src/semantic-highlighting.ts
```diff
@@ -2,14 +2,15 @@ import * as fs from 'fs';
...
-import * as vscodelc from 'vscode-languageclient';
+import * as vscodelc from 'vscode-languageclient/node';
 import * as vscodelct from 'vscode-languageserver-types';
 
 export function activate(client: vscodelc.LanguageClient,
                          context: vscode.ExtensionContext) {
   const feature = new SemanticHighlightingFeature(client, context);
   context.subscriptions.push(feature);
   client.registerFeature(feature);
+  client.registerProposedFeatures(); // enables LSP 3.16 semantic highlighting
```
hokein, line 13: I think this should be moved to `extension.ts`. This file is the legacy implementation, and is enabled by the "clangd.semanticHighlighting" flag.

Trass3r: Ah so it should always be enabled?

hokein: not really, it will be controlled by the vscode's builtin "Semantich Highlighting" flag (`"editor.semanticHighlighting.enabled"`)

Trass3r:
I'm not sure why the tests fail suddenly. They do work locally.

hokein:
> I'm not sure why the tests fail suddenly. They do work locally.

The CI is failing on the `src/type-hierarchy.ts` file.

Looks like the CI is running your PR on top of the master, but your branch seems behind the upstream. I think you need to rebase your branch to upstream master, and do the similar API changes for `src/type-hierarchy.ts` as well.

Trass3r:
Thanks, fixed.

hokein:
Thanks! looks good to me.

@sam-mccall do you have any other concern of this?

Trass3r:
I've been using it for a while, works nicely.

hokein:
Thanks, I believe it works. The only concern from our side is using the `next` version of `vscode-languageclient`, but it should be fine.

A last comment: there is a new release `7.0.0-next.9` of vscode-languageclient, it moves out the semanticToken from proposed state, could you update to that version as well? then we don't need to register proposed features in our extension.

Trass3r:
> Thanks, I believe it works.
You can check with the Inspect Editor Tokens tool.

> A last comment: there is a new release `7.0.0-next.9` of vscode-languageclient, it moves out the semanticToken from proposed state, could you update to that version as well? then we don't need to register proposed features in our extension.

Nice, done.

hokein (approved):
thanks!

Issue #68, opened 2020-09-09
"Open Type Hierarchy" menu should be hidden when clangd crashes


hokein:
Other menus ("go to definition" etc) are hidden when clangd crashes, I think we should align with them.

HighCommander4:
[Proposed fix](https://github.com/clangd/vscode-clangd/pull/73)

PR #73, opened 2020-09-20
Only show the 'Open Type Hierarchy' context menu item when the server…

HighCommander4:
… is running

Fixes https://github.com/clangd/vscode-clangd/issues/68

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
             "editor/context": [
                 {
                     "command": "clangd.typeHierarchy",
-                    "when": "resourceLangId == cpp",
+                    "when": "resourceLangId == cpp && extension.vscode-clangd.serverRunning",
                     "group": "0_navigation@4",
                     "_comment": "see https://github.com/microsoft/vscode-references-view/blob/f63eaed9934ca5ecc8f3fb3ca096f38c6e5e181f/package.json#L162"
                 }
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
             "editor/context": [
                 {
                     "command": "clangd.typeHierarchy",
-                    "when": "resourceLangId == cpp && extension.vscode-clangd.serverRunning",
+                    "when": "resourceLangId == cpp && extension.vscode-clangd.enableTypeHierarchy",
                     "group": "0_navigation@4",
                     "_comment": "see https://github.com/microsoft/vscode-references-view/blob/f63eaed9934ca5ecc8f3fb3ca096f38c6e5e181f/package.json#L162"
                 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -192,7 +192,7 @@
             "editor/context": [
                 {
                     "command": "clangd.typeHierarchy",
-                    "when": "resourceLangId == cpp",
+                    "when": "resourceLangId == cpp && extension.vscode-clangd.serverRunning",
```
hokein, line 195: This reminds me of a point I have missed in the previous review (sorry) -- we should only enable the typeHierarchy when the server capability `typeHierarchyProvider` is true, I think this should not be a big problem in practice (as typeHierarchy is implemented since clangd 9).

If the fix is trivial, would be nice to fix that -- I think we probably need a context value like `extension.vscode-clangd.hastypeHierarchyProvider`.

src/extension.ts
```diff
@@ -129,6 +129,15 @@ export async function activate(context: vscode.ExtensionContext) {
 
   const client = new ClangdLanguageClient('Clang Language Server',
                                           serverOptions, clientOptions);
+  client.onDidChangeState(stateChange => {
```
hokein, line 132: As this is only for type-hierarchy, I would use a specific name, and move this part of code to `type-hierarchy.ts` (to make it self-contained, it would be easier for us to delete it once there is an official implementation in VSCode).

As mentioned in another comment, if we can make `hastypeHierarchyProvider` work, we could reuse the `hastypeHierarchyProvider` as well.

package.json
HighCommander4: Added a check for `typeHierarchyProvider` in the updated patch.

src/extension.ts
HighCommander4: Refactored as suggested.

src/type-hierarchy.ts
```diff
@@ -82,6 +83,43 @@ class TypeHierarchyTreeItem extends vscode.TreeItem {
...
+    if (serverCapabilities.typeHierarchyProvider) {
+      this.serverSupportsTypeHierarchy = true;
+      this.recomputeEnableTypeHierarchy();
+    }
+  }
+
+  private recomputeEnableTypeHierarchy() {
+    if (this.state == vscodelc.State.Running) {
+      vscode.commands.executeCommand(
+          'setContext', 'extension.vscode-clangd.enableTypeHierarchy',
```
hokein, line 114: Looks like we are missing the `Starting` state enum, I think it is fine.

Do you know what's the default value of `extension.vscode-clangd.enableTypeHierarchy` if there is no ` vscode.commands.executeCommand` being executed yet? probably undefined?

hokein (commented):
Thanks, looks mostly good.

src/type-hierarchy.ts
HighCommander4: Yes, I expect it's undefined (and therefore falsy when evaluated as a boolean).

hokein (approved):
thanks, looks good.

PR #86, opened 2020-10-19
Add tree view for recent $/memoryUsage feature in clangd.

sam-mccall:


src/memory-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+  return {
+    title: title.substr(slash + 1), // display basename only
+    isFile: slash >= 0,
+    total: m._total,
+    self: m._self,
+    children: Object.keys(m)
+                  .filter(x => !x.startsWith('_'))
+                  .map(e => convert(m[e] as WireTree, e))
+                  .sort((x, y) => x.total - y.total)
+                  .reverse(),
```
kadircet, line 48: why not sort in the other way directly? (or why not sort based on title)

src/memory-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+  const feature = new ASTFeature(context);
+  context.client.registerFeature(feature);
+}
+
+// LSP wire format for this clangd feature.
+interface NoParams {}
+interface WireTree {
+  _self: number;
+  _total: number;
+  [child: string]: WireTree|number|string;
```
kadircet, line 24: my typescript isnt so strong, but why can this be anything but WireTree ? isn't it enough to handle `_self` and `_total` explicitly? (and even if that's not enough, where does the string come from)

src/memory-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+  private _onDidChangeTreeData = new vscode.EventEmitter<InternalTree|null>();
+  readonly onDidChangeTreeData = this._onDidChangeTreeData.event;
+
+  public getTreeItem(node: InternalTree): vscode.TreeItem {
+    const item = new vscode.TreeItem(node.title);
+    item.description = (node.total / 1024 / 1024).toFixed(2) + ' MB';
+    item.tooltip = `self=${node.self} total=${node.total}`;
+    if (node.isFile)
+      item.iconPath = new vscode.ThemeIcon('symbol-file');
+    else if (!node.children.length)
```
kadircet, line 95: do we intentionally not set any icons for non-leaf (and non-file) nodes?

src/memory-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+      if (node.children.length >= 6 || node.isFile)
+        item.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+      else
+        item.collapsibleState = vscode.TreeItemCollapsibleState.Expanded;
+    }
+    return item;
+  }
+
+  public async getChildren(t?: InternalTree): Promise<InternalTree[]> {
+    return t ? t.children : this.root ? [this.root] : [];
```
kadircet, line 107: i bet this reads correctly, but might be nice to have some parentheses to make it explicit :D

src/memory-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// LSP wire format for this clangd feature.
+interface NoParams {}
+interface WireTree {
+  _self: number;
+  _total: number;
+  [child: string]: WireTree|number|string;
+}
+export const RequestType =
```
kadircet, line 26: nit: maybe rename to `MemoryUsageRequest` ?

src/memory-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+    self: m._self,
+    children: Object.keys(m)
+                  .filter(x => !x.startsWith('_'))
+                  .map(e => convert(m[e] as WireTree, e))
+                  .sort((x, y) => x.total - y.total)
+                  .reverse(),
+  };
+}
+
+class ASTFeature implements vscodelc.StaticFeature {
```
kadircet, line 52: s/ASTFeature/MemoryUsageFeature/

kadircet (approved):
thanks, LGTM

src/memory-usage.ts
sam-mccall: Correct. This is subjective, but I couldn't find an icon that seemed to add signal rather than just visual noise.

src/memory-usage.ts
sam-mccall: _self and _total have to conform to this constraint too, unfortunately. I couldn't find a way to more precisely express the type. This is one of the reasons for mapping quickly to another type :-)

(There's a workaround involving intersection types, but they must be `type` rather than `interface` and `type`s can't be recursive as they're just aliases)

src/memory-usage.ts
sam-mccall: You're right. The intention was to sort by [total, title] but I messed it up.

The code to sort by a compound key is awkward in JS, but sort() is stable, so we sort the titles first and then the totals.

hokein:
might be worth to mention this in https://clangd.llvm.org/extensions.html

PR #87, opened 2020-10-19
Use built-in icons instead of SVGs for type hierarchy.

sam-mccall:
I was experimenting with more tree views (e.g. #86) and ran into this simplification.

The built-in icons are shown here: https://code.visualstudio.com/api/references/icons-in-labels
We could also consider chevron-{up,down} or arrow-{up,down}


HighCommander4 (approved):
Thanks, good find!

PR #96, opened 2020-11-09
Save the Disposable returned when registering clangd.memoryUsage

### LINKED ISSUE #95: Command 'clangd.restart' tries to re-register 'clangd.memoryUsage'?
I wanted to try out the "watch `compile_commands.json`" enhancement added by #33, so I cloned to commit 0b171de, and tried it on Debian 10.4 with VSCode 1.51.0 and clangd 11.0.1 installed from the "llvm-toolchain-buster-11" repository.

Changes to `compile_commands.json` result in a pop-up, as expected, but when I select "restart", the `clangd` process exits but silently fails to restart, with no hints in the "Clangd Language Server" log. Further investigation revealed it to be a problem with the "clangd.restart" command itself: in the v0.1.7 release, this works fine, but in commit 0b171de it produces this error:

`Command 'Restart the clangd language server' resulted in an error (command 'clangd.memoryUsage' already exists)`

I could hackily make this error go away and get the behaviour intended by #33 by commenting out the lines in `out/src/memory-usage.js`:

`vscode.commands.registerCommand('clangd.memoryUsage', ...);`
`vscode.commands.registerCommand('clangd.memoryUsage.close', ...);`

I suspect you just need to check whether that command has already been registered, but my TypeScript knowledge is almost nonexistent so I'm not going to embarrass myself by raising a PR for it!

makestuff:
Attempt to fix #95.

### DIFF
--- a/src/memory-usage.ts
+++ b/src/memory-usage.ts
@@ -57,13 +57,14 @@ class MemoryUsageFeature implements vscodelc.StaticFeature {
             'setContext', 'clangd.memoryUsage.hasData', adapter.root != null));
     this.context.subscriptions.push(
         vscode.window.registerTreeDataProvider('clangd.memoryUsage', adapter));
-    vscode.commands.registerCommand('clangd.memoryUsage', async () => {
-      const usage =
-          await this.context.client.sendRequest(MemoryUsageRequest, {});
-      adapter.root = convert(usage, '<root>');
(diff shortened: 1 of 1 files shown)


hokein (approved):
Thanks for spotting and fixing it, particularly near the new release time!

PR #97, opened 2020-11-12
Restrict memory usage button to its own view

trond-snekvik:
The "show memory usage" button did not have a when clause, making it
show up in every view title across the editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,6 +238,7 @@
                 },
                 {
                     "command": "clangd.memoryUsage",
+                    "when": "view == clangd.memoryUsage",
                     "group": "navigation"
                 }
             ],


Trass3r (approved):

hokein:
Thanks for the patch!

Can you paste a screenshot to show the problem? I'm not sure I understand the issue.

Trass3r:
For me it showed up in the test explorer yesterday.

trond-snekvik:
> Can you paste a screenshot to show the problem? I'm not sure I understand the issue.

Sure!

Here the button shows up in the problems panel title menu:
clangd_problems

And here it is showing up in an explorer tree view owned by another extension:

clangd_devicetree

Both instances are gone after this patch.

hokein:
oh, I see, that makes sense. Thanks again!

hokein (approved):

PR #99, opened 2020-11-14
Add debouncing mechanism to avoid restarting clangd too many times

### LINKED ISSUE #98: Error restarting server when using onConfigChanged
When using onConfigChanged and the compile_commands.json file gets updated I get an error when the extension tries to restart the server

image

**System information**
Clangd version: trunk
clangd extension version: 0.18
Operating system: Ubuntu 20.04

rapgenic:
Fixes #98

### DIFF
--- a/src/config-file-watcher.ts
+++ b/src/config-file-watcher.ts
@@ -3,6 +3,8 @@ import * as vscode from 'vscode';
 import {ClangdContext} from './clangd-context';
 import * as config from './config';
 
+const RESTART_DEBOUNCE = 2000;
+
 export function activate(context: ClangdContext) {
   if (config.get<string>('onConfigChanged') != 'ignore') {
     const watcher = new ConfigFileWatcher(context);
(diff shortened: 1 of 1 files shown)


njames93:
@Trass3r Do you want to give this a try, It fixed it on my end but there could be something subtly different on your end.

src/config-file-watcher.ts
```diff
@@ -11,6 +13,7 @@ export function activate(context: ClangdContext) {
 
 class ConfigFileWatcher {
   private databaseWatcher: vscode.FileSystemWatcher = undefined;
+  private pendingRestartTimer: NodeJS.Timer = undefined;
```
hokein, line 16: maybe name it `debounceTimer`?

src/config-file-watcher.ts
```diff
@@ -26,12 +29,23 @@ class ConfigFileWatcher {
...
 
+  async debouncedHandleConfigFilesChanged(uri: vscode.Uri) {
+    if (this.pendingRestartTimer) {
+      clearTimeout(this.pendingRestartTimer);
+    }
+
+    this.pendingRestartTimer = setTimeout(async () => {
+      await this.handleConfigFilesChanged(uri);
+      this.pendingRestartTimer = undefined;
+    }, RESTART_DEBOUNCE);
```
hokein, line 46: nit: just inline the value here since we use it only once.

rapgenic:
Done! 

I also updated with the new changes on master

hokein (approved):
Thanks.

PR #101, opened 2020-11-18
Prevent file watcher error when no folder is open

rapgenic:
This prevents an error that occurs when the extension tries to register the watcher, but no folder is open (for example when only a single c file is).

With this fix, in this case no watcher is registered.


hokein (approved):
Thanks, this looks good.

The CI failure is irrelevant to this patch. I'll fix that.

PR #102, opened 2020-11-20
Naming cleanups

sam-mccall:
Log in output is named "clangd" (was "Clangd Language Server")

Commands shown in palette are consistently "clangd: foo bar" (some were "foo clangd bar")
Exception: type hierarchy (proposed as standard LSP feature)

Internal resource IDs consistently clangd.foo (some were extension.vscode-clangd.foo)

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,23 +144,23 @@
         "commands": [
             {
                 "command": "clangd.switchheadersource",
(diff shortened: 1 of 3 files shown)


hokein (approved):
btw, I'm fixing the ci now.

PR #103, opened 2020-11-20
Remove the deprecated set-env command.

hokein:
See https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


hokein:
The CI is happy now.

PR #104, opened 2020-11-20
Add "show AST" command, based on textDocument/ast extension.

sam-mccall:
See https://github.com/llvm/llvm-project/commit/8adc4d1ec76471bc283d888f3077f7d8f591d6ad
This extension is new, so the feature doesn't work with clangd 11 and earlier.

2020-11-20-134221_1126x790_scrot

Implementation is pretty similar to memoryusage (though a bit simpler I think)

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,6 +196,16 @@
                 "command": "clangd.memoryUsage.close",
                 "title": "Close",
                 "icon": "$(panel-close)"
+            },
+            {
+                "command": "clangd.ast",
+                "title": "Show AST",
+                "icon": "$(list-tree)"
+            },
+            {
+                "command": "clangd.ast.close",
+                "title": "Close",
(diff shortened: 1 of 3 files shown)


kadircet (approved):
As a general point; it might be nice to have some pointers for the possible values of "roles" and "kinds". Currently it feels like magic when reading the mappings. I suppose it is OK for now, but it might be a hard-to-discover concept for outsiders.

sam-mccall:
> As a general point; it might be nice to have some pointers for the possible values of "roles" and "kinds". Currently it feels like magic when reading the mappings. I suppose it is OK for now, but it might be a hard-to-discover concept for outsiders.

Yeah, added a link to the source code and repeated a bit what these are.

It's a bit weird and coupled, but it works and I can't think of a better alternative. (I don't think "no icons" is actually better!)

PR #106, opened 2020-11-22
Remove the file watch on .clang-tidy

njames93:
Restarting the server when this file changes doesn't make that much sense.
There can be .clang-tidy files in sub directories(and parent directories) that wont be watched so having the server restart when only the base directory configuration changes leads to non deterministic behaviour.
Ensuring clang-tidy configurations are up to date should be done on the clangd server instead. This could be done in a similar way to how the config files are cached but validated against modification time when accessed.
Going this route will also work for other editors that use clangd.


hokein (approved):
thanks. I noticed this and didn't think hard of it during the review of file-watcher patch.

I think this makes sense, only watching base-directory `.clang-tidy` is marginally useful, so LGTM.

PR #108, opened 2020-11-24
Add a restartAfterCrash option

hokein:
Allowing us to control the maximum clangd restart times (by default it is 4), this would make the life of tracking/reporting crashes a bit easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,11 @@
                     "default": true,
                     "description": "Always rank completion items on the server as you type. This produces more accurate results at the cost of higher latency than client-side filtering."
                 },
+                "clangd.maxRestartCount": {
+                    "type": "number",
+                    "default": 4,
+                    "description": "The maximum count that the extension will restart clangd if clangd crashes."
+                },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -124,6 +124,11 @@
                     "default": true,
                     "description": "Always rank completion items on the server as you type. This produces more accurate results at the cost of higher latency than client-side filtering."
                 },
+                "clangd.maxRestartCount": {
```
sam-mccall, line 127: as discussed offline...

I think from a UX perspective "do you want to restart" is a much easier question to answer, and being able to set the count to 2 or 99 isn't actually very important.

So I'd make this "restartAfterCrash", boolean, default true, and the logic could be restartAfterCrash ? 4 : 0 or something like that.

sam-mccall (approved):
This make a lot of sense for those of us that really want to track down/report crashes.

I think we can simplify the option a little though.

package.json
hokein: Thanks, done.

PR #141, opened 2021-01-29
Warn about conflict with ms-vscode.cpptools

kirillbobyrev:
`vscode-clangd` is known to create confusing user experience when `ms-vscode.cpptools` extension is enabled but it is not clear to the users. This patch creates a warning for described scenario.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -13,13 +13,6 @@ export async function activate(context: vscode.ExtensionContext) {
   const clangdContext = new ClangdContext;
   context.subscriptions.push(clangdContext);
 
-  if (vscode.extensions.getExtension('ms-vscode.cpptools') !== undefined) {
-    vscode.window.showWarningMessage(
-        'You have Microsoft C++ (ms-vscode.cpptools) extension enabled, it is ' +
-            'known to conflict with vscode-clangd. We recommend disabling it.',
-        'Got it');
-  }
-
   // An empty place holder for the activate command, otherwise we'll get an
   // "command is not registered" error.
   context.subscriptions.push(
@@ -31,4 +24,28 @@ export async function activate(context: vscode.ExtensionContext) {
       }));
 
   await clangdContext.activate(context.globalStoragePath, outputChannel);
+
+  setInterval(function() {
+    const cpptools = vscode.extensions.getExtension('ms-vscode.cpptools');
+    if (cpptools !== undefined && cpptools.isActive) {
+      const intellisenseEnabled =
+          vscode.workspace.getConfiguration('C_Cpp').get('intelliSenseEngine');
+      const DisableIt = 'Disable cpptools';
+      if (intellisenseEnabled === 'Default' || intellisenseEnabled === true) {
+        vscode.window
+            .showWarningMessage(
+                'You have Microsoft C++ (ms-vscode.cpptools) extension enabled, it is ' +
+                    'known to conflict with vscode-clangd. We recommend disabling it.',
+                DisableIt, 'Got it')
+            .then(selection => {
+              console.log('selection');
+              if (selection === DisableIt) {
+                vscode.workspace.getConfiguration('C_Cpp').update(
+                    'intelliSenseEngine', false,
+                    vscode.ConfigurationTarget.Global);
+              }
+            });
+      }
+    }
+  }, 5000);
 }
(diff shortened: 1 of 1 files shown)


rapgenic:
Maybe would be even more helpful to add another button saying "Disable it" that writes 
```
    "C_Cpp.intelliSenseEngine": "Disabled",
```
in the config?

kirillbobyrev:
> Maybe would be even more helpful to add another button saying "Disable it" that writes
> 
> ```
>     "C_Cpp.intelliSenseEngine": "Disabled",
> ```
> 
> in the config?

Good idea! Am I correct to think this doesn't disable the whole extension altogether? I'm a bit confused by the documentation of that option: IIUC Intellisense itself is the completion provider, so this should just disable the completion source, am I correct? Would that be enough to do that in order to get around the problems we currently have with both extensions enabled?

rapgenic:
It should disable language server features (see https://github.com/microsoft/vscode-cpptools/issues/4979), which however are completely taken over by clangd; the other features, such as debug and build keep working.

In my opinion it shoud be more than enough to avoid the compatibility issues, I myself have been using that option to keep the debug support, while using clangd for the language server features.

kirillbobyrev:
The detection of mstools being active and the respective setting being either default or `true` is OK but the following doesn't work as expected:

```typescript
            .then(selection => {
              console.log('selection');
              if (selection === DisableIt) {
                vscode.workspace.getConfiguration('C_Cpp').update(
                    'intelliSenseEngine', false,
                    vscode.ConfigurationTarget.Global);
              }
            })
```

I can't really see the global config being file being updated. And I also can't get into the function from the debugger for some reason though I'm not experienced with TS and debugging it at all.

Am I trying to update it wrong?

kirillbobyrev:
nvm, apparently the labels & order were behaving weirdly, changing labels makes everything work fine. I removed "Got it" button since it's essentially the same as "close warning message" and only have the one which provides the needed function.

This patch works now, ready for a review.

src/extension.ts
```diff
@@ -24,4 +24,24 @@ export async function activate(context: vscode.ExtensionContext) {
       }));
 
   await clangdContext.activate(context.globalStoragePath, outputChannel);
+
+  setInterval(function() {
+    const cpptools = vscode.extensions.getExtension('ms-vscode.cpptools');
+    if (cpptools !== undefined && cpptools.isActive) {
```
sam-mccall, line 30: nit: maybe just cppTools && cppTools.isActive? (the docs say it'll be undefined rather than null but...)

src/extension.ts
```diff
@@ -24,4 +24,24 @@ export async function activate(context: vscode.ExtensionContext) {
...
+      const cpptoolsConfiguration = vscode.workspace.getConfiguration('C_Cpp');
+      const cpptoolsEnabled = cpptoolsConfiguration.get('intelliSenseEngine');
+      if (cpptoolsEnabled !== 'Disabled') {
+        vscode.window
+            .showWarningMessage(
+                'You have Microsoft C++ (ms-vscode.cpptools) extension ' +
```
sam-mccall, line 36: We should be careful about the wording here, as many users won't have a sophisticated understanding of plugins vs features of plugins etc, but we expect them to be able to undo this change if needed.
I think we need to call out that we're only disabling IntelliSense here.

Maybe:

You have both the Microsoft C++ (cpptools) extension and clangd extension enabled.
The Microsoft IntelliSense features conflict with clangd's code completion, diagnostics etc.
[Disable IntelliSense]

sam-mccall (approved):

xpinguin:
Hi!

Could you please suggest a way to either disable this annoying warning (along with the automatic `cpptools` switch off), or perhaps give me some trivial advice on the ways to edit the published extension locally?

I'm new to VSCode, using it mostly for C or C++ projects (don't like spending a lot of time configuring clangd, no incremental parsing, etc.) - my workflow with both extensions enabled was fluent and, contrary to the PR statement, quite the opposite of the word "confusing"...

Would appreciate any help, thanks in advance!

HighCommander4:
@xpinguin Maybe file a new issue suggesting to add an option to disable the warning?

As for editing the extension locally, the only method I'm familiar with is to clone the repo, run `npm install`, make the edit, run `npm run package` and then install the resulting `.vsix` file using the vscode command "Extensions: Install from VSIX...".

xpinguin:
Neat! My appreciations, that solved my problem. Sure, I hope I'll create an issue when weekends would (finally!) arrive :)

PR #150, opened 2021-02-18
Upgrade vscode-languageclient to version 7.1.0-next.1

HighCommander4:
This picks up the fix for
https://github.com/microsoft/vscode-languageserver-node/issues/576
which fixes an issue where semantic highlighting flickers.

The patch also makes a corresponding update to API usage.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,2161 @@
 {
     "name": "vscode-clangd",
     "version": "0.1.10",
-    "lockfileVersion": 1,
+    "lockfileVersion": 2,
     "requires": true,
+    "packages": {
+        "": {
+            "version": "0.1.10",
+            "license": "MIT",
+            "dependencies": {
+                "@clangd/install": "0.1.3",
+                "abort-controller": "^3.0.0",
+                "jsonc-parser": "^2.1.0",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,8 +1,2161 @@
 {
     "name": "vscode-clangd",
     "version": "0.1.10",
-    "lockfileVersion": 1,
+    "lockfileVersion": 2,
```
hokein, line 4: is this change intentional? Looking at our previous upgrade PR, it didn't change the version, e.g. https://github.com/clangd/vscode-clangd/pull/91/files

HighCommander4: The change was made automatically when I ran `npm install` after updating the dependency in `package.json`.

In terms of what it means, a quick look at the [docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) suggests that:

  * npm v7 introduced a new lockfile format version (version 2)
  * version 2 is backwards compatible with version 1

So I think this should be fine.

hokein: I see, fair enough. Thanks for the explanations!

hokein (approved):
thanks!

btw, @kirillbobyrev has setup a bot to do this kind of update automatically.

PR #153, opened 2021-02-20
Ensure clangd is executed in vscode.workspace.rootPath

allight:
If the rootPath of a project is a symlink vscode can execute clangd in
the users home-directory. This can cause problems if one is, for
example, using a wrapper script which relies on knowing the current
project. This solves this issue by forcing the clangd to execute in
`vscode.workspace.rootPath` if available.

Test: manual

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -50,9 +50,15 @@ export class ClangdContext implements vscode.Disposable {
     if (!clangdPath)
       return;
 
+    const rootPath: string = vscode.workspace.rootPath
+    const targetCwd: string = (rootPath !== undefined) ? rootPath : process.cwd();
+    const clangdExecOptions: vscodelc.ExecutableOptions = {
+      cwd: targetCwd
+    }
     const clangd: vscodelc.Executable = {
       command: clangdPath,
-      args: config.get<string[]>('arguments')
+      args: config.get<string[]>('arguments'),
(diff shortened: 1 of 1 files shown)


allight:
Just for background this is an issue I personally ran into since my job has a highly patched version of clangd in the path which breaks on some of the projects I work on. To work around this I have a script that looks at the current project and execs the appropriate clangd set as the global `clangd.path`. It uses the PWD to decide which clangd to use so this weird behavior with vscode not wanting to chdir into a symlink bit me by causing the wrapper to get confused since it was being run in $HOME.

src/clangd-context.ts
```diff
@@ -50,9 +50,14 @@ export class ClangdContext implements vscode.Disposable {
     if (!clangdPath)
       return;
 
+    const rootPath: string = vscode.workspace.rootPath;
+    const targetCwd: string =
```
sam-mccall, line 54: nit: I guess we don't need three variables here, can we just inline as

```
  options: {cwd: vscode.workspace.rootPath || process.cwd() }
```

I don't think this is too cryptic to read
or just initialize directly as vscode.workspace.rootPath || process.cwd()

sam-mccall (approved):
Thanks!

src/clangd-context.ts
allight: done

PR #160, opened 2021-03-02
Prompt when workspace overrides clangd.path and clangd.arguments

sam-mccall:


hokein (approved):
Thanks, looks pretty solid. I also test it locally.

adzenith:
Is there any way to reset this if you accidentally click "no" at the prompt?

sam-mccall:
We save the answer for the latest value seen of each setting. So you can change the setting in workspace/.vscode/settings.json and reload the window to get re-prompted. (And then, change it back!)

simon8233:
is it any way to disable it? 
like $HOME/.clangd_config
we use clangd.path and clangd.arguments and remote index server setup as one workspace file in internal.

HighCommander4:
@simon8233 it might be easier for the question to be noticed if you file a new issue for it

simon8233:
Thanks for your feedback. I will raise question for this.

PR #178, opened 2021-04-18
Enable "strict" compiler option

### LINKED ISSUE #61: Should build with --strict
To catch bugs early.
https://github.com/clangd/vscode-clangd/blob/6f66a43d80427becc3560418b0b203dfaf09d1ee/tsconfig.json#L19
```json
        "strict": true,
```

BigBahss:
Closes #61.

Changes have been made to adhere to strict type checking:
- Non-null assertions and definite-assignment assertions (`!` post-fix operator) have been added to please the compiler (these do not affect the compiled output). These assertions are not always strictly "safe", but are not any less safe than not having strict type checking at all. Plus, their presence in the code can be a signal that maybe the design should be changed to safely ensure that the types are not null/undefined.
- Some optional chaining (`?.`) has been been added when properties can be null/undefined.
- Some incorrect uses of `null` have been replaced with `undefined`.
- Add `defaultValue` arguments to `WorkspaceConfiguration.get()` so that it will not return undefined. The default values being passed are the default values defined in the package.json. Realistically, the function should not return undefined when default values are defined in the package.json, but this pleases the compiler.

The changes I've made were aimed at not changing the behavior of the extension, and weren't necessarily aimed at fixing any potential bugs. Strict type checking will make it less likely to introduce new bugs in future changes.

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -24,8 +24,7 @@ interface ASTNode {
   range?: vscodelc.Range;
 }
 const ASTRequestType =
-    new vscodelc.RequestType<ASTParams, ASTNode|undefined, void>(
-        'textDocument/ast');
+    new vscodelc.RequestType<ASTParams, ASTNode|null, void>('textDocument/ast');
 
 class ASTFeature implements vscodelc.StaticFeature {
   constructor(private context: ClangdContext) {
@@ -41,10 +40,12 @@ class ASTFeature implements vscodelc.StaticFeature {
         adapter.onDidChangeTreeData((_) => {
           vscode.commands.executeCommand('setContext', 'clangd.ast.hasData',
                                          adapter.hasRoot());
+          // Work around https://github.com/microsoft/vscode/issues/90005
           // Show the AST tree even if it's been collapsed or closed.
           // reveal(root) fails here: "Data tree node not found".
           if (adapter.hasRoot())
-            tree.reveal(null!);
+            // @ts-ignore
+            tree.reveal(null);
         }),
         vscode.window.registerTreeDataProvider('clangd.ast', adapter),
         // Create the "Show AST" command for the context menu.
@@ -62,7 +63,7 @@ class ASTFeature implements vscodelc.StaticFeature {
               if (!item)
                 vscode.window.showInformationMessage(
                     'No AST node at selection');
-              adapter.setRoot(item, editor.document.uri);
+              adapter.setRoot(item ?? undefined, editor.document.uri);
             }),
         // Clicking "close" will empty the adapter, which in turn hides the
         // view.
@@ -110,8 +111,8 @@ const KindIcons: {[type: string]: string} = {
 function describe(role: string, kind: string): string {
   // For common roles where the kind is fairly self-explanatory, we don't
   // include it. e.g. "Call" rather than "Call expression".
-  if (role == 'expression' || role == 'statement' || role == 'declaration' ||
-      role == 'template name')
+  if (role === 'expression' || role === 'statement' || role === 'declaration' ||
+      role === 'template name')
     return kind;
   return kind + ' ' + role;
 }
@@ -121,7 +122,7 @@ class TreeAdapter implements vscode.TreeDataProvider<ASTNode> {
   private root?: ASTNode;
   private doc?: vscode.Uri;
 
-  hasRoot(): boolean { return this.root != null; }
+  hasRoot(): boolean { return this.root !== undefined; }
 
   setRoot(newRoot: ASTNode|undefined, newDoc: vscode.Uri|undefined) {
     this.root = newRoot;
@@ -165,11 +166,11 @@ class TreeAdapter implements vscode.TreeDataProvider<ASTNode> {
   }
 
   public getParent(node: ASTNode): ASTNode|undefined {
-    if (node == this.root)
+    if (node === this.root)
       return undefined;
     function findUnder(parent: ASTNode|undefined): ASTNode|undefined {
-      for (const child of parent?.children || []) {
-        const result = (node == child) ? parent : findUnder(child);
+      for (const child of parent?.children ?? []) {
+        const result = (node === child) ? parent : findUnder(child);
         if (result)
           return result;
       }
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -24,7 +24,8 @@ interface ASTNode {
   range?: vscodelc.Range;
 }
 const ASTRequestType =
-    new vscodelc.RequestType<ASTParams, ASTNode|null, void>('textDocument/ast');
+    new vscodelc.RequestType<ASTParams, ASTNode|undefined, void>(
+        'textDocument/ast');
 
 class ASTFeature implements vscodelc.StaticFeature {
   constructor(private context: ClangdContext) {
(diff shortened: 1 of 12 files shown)


src/ast.ts
```diff
@@ -39,11 +40,11 @@ class ASTFeature implements vscodelc.StaticFeature {
...
         adapter.onDidChangeTreeData((_) => {
           vscode.commands.executeCommand('setContext', 'clangd.ast.hasData',
-                                         adapter.hasRoot())
-          // Show the AST tree even if it's beet collapsed or closed.
+                                         adapter.hasRoot());
+          // Show the AST tree even if it's been collapsed or closed.
           // reveal(root) fails here: "Data tree node not found".
           if (adapter.hasRoot())
-          tree.reveal(null);
+            tree.reveal(null!);
```
BigBahss, line 47: This one is kind of silly since we are asserting that null is not null. But, the API expects a non-null object, and if passing null works around the issue described, then this should be fine.

src/semantic-highlighting.ts
```diff
@@ -292,7 +292,7 @@ export class ThemeRuleMatcher {
   // Returns the best rule for a scope.
   getBestThemeRule(scope: string): TokenColorRule {
     if (this.bestRuleCache.has(scope))
-      return this.bestRuleCache.get(scope);
+      return this.bestRuleCache.get(scope)!;
```
BigBahss, line 295: I would consider this a safe non-null assertion, since we have already checked that the `Map` has the value before getting it.

src/ast.ts
```diff
@@ -24,7 +24,8 @@ interface ASTNode {
   range?: vscodelc.Range;
 }
 const ASTRequestType =
-    new vscodelc.RequestType<ASTParams, ASTNode|null, void>('textDocument/ast');
+    new vscodelc.RequestType<ASTParams, ASTNode|undefined, void>(
```
sam-mccall, line 27: I don't understand this one, the JSON-RPC response may literally be `null`

src/ast.ts
sam-mccall: I think maybe we should just @ts-ignore here instead, the current plan is for them to fix the typings: https://github.com/microsoft/vscode/commit/e95c40c1e461f83f5ff1c68d61253ab23c85f86b

src/ast.ts
```diff
@@ -163,16 +164,16 @@ class TreeAdapter implements vscode.TreeDataProvider<ASTNode> {
     return element.children || [];
   }
 
-  public getParent(node: ASTNode): ASTNode {
+  public getParent(node: ASTNode): ASTNode|undefined {
```
sam-mccall, line 167: There are a bunch of places in this class where you've changed `null` to `undefined` - is there some fundamental reason why null is incorrect? (It makes sense to be consistent within a module of course.

My understanding is this is mostly a stylistic choice, and my preference would be to use null for "deliberately" null values, but this isn't my primary language...

src/clangd-context.ts
```diff
@@ -58,7 +58,7 @@ export class ClangdContext implements vscode.Disposable {
           await config.getSecureOrPrompt<string[]>('arguments', workspaceState),
       options: {cwd: vscode.workspace.rootPath || process.cwd()}
     };
-    const traceFile = config.get<string>('trace');
+    const traceFile = config.get<string>('trace', '');
```
sam-mccall, line 61: I think that echoing defaults already given in the package.json is unacceptable duplication (and causes practical problems e.g. someone changing the default in the wrong place).

Can we add a `!` assertion in config.get() instead?

src/config.ts
```diff
@@ -17,7 +18,7 @@ export function getSecure<T>(key: string, workspaceState: vscode.Memento): T|
 // Like get(), but won't implicitly load settings from workspace config.
 // If there is workspace config, prompts the user and caches the decision.
 export async function getSecureOrPrompt<T>(
-    key: string, workspaceState: vscode.Memento): Promise<T> {
+    key: string, workspaceState: vscode.Memento): Promise<T|undefined> {
```
sam-mccall, line 21: (again here I think we should assert the result rather than returning a maybe-result)

src/install.ts
```diff
@@ -21,8 +21,9 @@ export async function activate(
       'clangd.install', async () => common.installLatest(ui)));
   context.subscriptions.push(vscode.commands.registerCommand(
       'clangd.update', async () => common.checkUpdates(true, ui)));
-  const status = await common.prepare(ui, config.get<boolean>('checkUpdates'));
-  return status.clangdPath;
+  const status =
```
sam-mccall, line 24: The function comment says "returns the path or null", so I think we return status?.clangdPath instead maybe, and change the signature?

src/memory-usage.ts
```diff
@@ -80,8 +80,8 @@ class MemoryUsageFeature implements vscodelc.StaticFeature {
 class TreeAdapter implements vscode.TreeDataProvider<InternalTree> {
   private root_?: InternalTree;
 
-  get root(): InternalTree|null { return this.root_; }
-  set root(n: InternalTree|null) {
+  get root(): InternalTree|undefined { return this.root_; }
```
sam-mccall, line 83: again why these null->undefined changes?

tsconfig.json
```diff
@@ -16,7 +16,7 @@
         ],
         "sourceMap": true,
         "rootDir": ".",
-        "alwaysStrict": true,
+        "strict": true,
```
sam-mccall, line 19: It's surprising that "strict" is more strict than "alwaysStrict" :-)

sam-mccall (commented):
Thanks for doing this cleanup! And sorry if some of these comments are dumb, I'm not deeply familiar with TS's type system. I do think if we're going to enable strict mode then I should understand it though...

src/clangd-context.ts
BigBahss: That's fair. I believe the defaults can be retrieved through `WorkspaceConfiguration.inspect()`, but that may be more trouble than it's worth. Realistically adding asserts here would only cause a problem if a default value was removed from `package.json`, but I expect that's quite unlikely.

src/install.ts
BigBahss: Good catch, the caller of this function already checks if the result is falsy. I'll fix that and add `null` to the return type.

src/memory-usage.ts
BigBahss: Addressed in a previous comment, but it's because `private root_?: InternalTree` is already declared as being `InternalTree | undefined`.

tsconfig.json
BigBahss: Yeah, it really seems like a "gcc -Wall" scenario haha.

BigBahss (commented):
Addressed comments (further consensus is needed on the `null` vs `undefined` issue).

src/ast.ts
sam-mccall: What I mean is, I think this type param corresponds to the actual JSON-RPC data sent by clangd, which definitely can be `null` (there's no `undefined` in JSON!).

So I think ASTRequestType needs to use `null` for correctness, but it's fine to convert that to `undefined` immediately after calling `sendRequest`.

src/ast.ts
sam-mccall: No worries at all. I think if `undefined` is privileged by the `?.` operators etc then ergonomics is a good reason to prefer it.
(And it seems so, playground says `(null as any)?.foo` is `undefined`)
And this seems to be MS's preference, and it's mostly their APIs we're dealing with.

BigBahss: I agree. So just to confirm, are deciding to use `undefined` for values stored by the extension and accepting `null` for incoming values (when necessary)?

src/ast.ts
BigBahss: I was not aware of that. Then I agree, this should be changed back to `null` for the RequestType.

src/config.ts
BigBahss: I'll fix that.

src/ast.ts
sam-mccall: That sounds good to me!

BigBahss:
I was going crazy there for a minute wondering how the CI was failing on a non-existent file when the tests passed on my machine 😅. Anyways I merged and modified the new file for strict type checking.

I fixed all the things that were mentioned. Additionally I changed all `==` and `!=` to `===` and `!==` since that's very closely related to strict type checking. Let me know if I missed anything or whether you have further concerns.

sam-mccall (approved):
Thanks a lot!

BigBahss:
Happy to help 👍

PR #181, opened 2021-04-22
Add commands to open global and project config

### LINKED ISSUE #176: add command open clangd configure
as clangd support configure file, can we have a way open/create configure file more quickly?

https://clangd.llvm.org/config.html

njames93:
Global config opening will likely not work on remote setups and is as of yet untested on windows and mac builds.

Fixes #176

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,7 +1,6 @@
 import * as vscode from 'vscode';
 
 import {ClangdContext} from './clangd-context';
-import * as OpenConfig from './open-config';
 
 /**
  *  This method is called when the extension is activated. The extension is
@@ -24,8 +23,6 @@ export async function activate(context: vscode.ExtensionContext) {
         await clangdContext.activate(context.globalStoragePath, outputChannel,
                                      context.workspaceState);
       }));
-
-  OpenConfig.activate(context);
 
   await clangdContext.activate(context.globalStoragePath, outputChannel,
                                context.workspaceState);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,6 +1,51 @@
+import * as os from 'os'
+import * as path from 'path'
 import * as vscode from 'vscode';
 
 import {ClangdContext} from './clangd-context';
+
+function getConfigFolder(): string {
+  switch (os.platform()) {
+  case 'win32':
+    if (process.env.LocalAppData)
+      return process.env.LocalAppData;
+    break;
+  case 'darwin':
+    if (process.env.HOME)
+      return path.join(process.env.HOME, 'Library', 'Preferences');
+    break;
+  case 'linux':
+    if (process.env.XDG_CONFIG_HOME)
+      return process.env.XDG_CONFIG_HOME;
+    if (process.env.HOME)
+      return path.join(process.env.HOME, '.config');
+    break;
+  default:
+    break;
+  }
+  return null;
+}
+
+function getConfigFile(): string {
+  const Folder = getConfigFolder();
+  if (!Folder)
+    return null;
+  return path.join(Folder, 'clangd', 'config.yaml');
+}
+
+async function openConfigFile(path: vscode.Uri) {
+  let p = path;
+  try {
+    await vscode.workspace.fs.stat(path);
+  } catch {
+    // File doesn't exist, create a scratch file.
+    p = path.with({scheme: 'untitled'});
+  }
+  vscode.workspace.openTextDocument(p).then((a => {
+    vscode.languages.setTextDocumentLanguage(a, 'yaml');
+    vscode.window.showTextDocument(a);
+  }));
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -24,6 +69,28 @@ export async function activate(context: vscode.ExtensionContext) {
                                      context.workspaceState);
       }));
 
+  // Create a command to open the project root .clangd configuration file.
+  context.subscriptions.push(
+      vscode.commands.registerCommand('clangd.projectSettings', () => {
+        if (vscode.workspace.workspaceFolders) {
+          const folder = vscode.workspace.workspaceFolders[0];
```
sam-mccall, line 76: Hmm, how sure are we that this is the right directory (that workspace folders correspond to the real project root, and folder 0 is the right one?)

src/extension.ts
```diff
@@ -24,6 +69,28 @@ export async function activate(context: vscode.ExtensionContext) {
                                      context.workspaceState);
       }));
 
+  // Create a command to open the project root .clangd configuration file.
+  context.subscriptions.push(
+      vscode.commands.registerCommand('clangd.projectSettings', () => {
+        if (vscode.workspace.workspaceFolders) {
```
sam-mccall, line 75: can we be a bit more explicit about these conditions? I think we're really testing length == 0 here, right?

sam-mccall (commented):
Thanks for putting this together!
I'm not 100% sure that we have enough info to decide where the project config should go, but maybe my vscode usage is atypical. @hokein @kbobyrev WDYT about writing .clangd to workspace folders?

src/extension.ts
njames93: There's unfortunately not a great way of going about this. In a single workspace mode this should be sufficient. But once you start working with multi folder workspaces things start getting a little hazy. However a lot of projects users are likely to work on only have 1 root folder so this should be helpful for the majority of users.

hokein:
> Thanks for putting this together!
> I'm not 100% sure that we have enough info to decide where the project config should go, but maybe my vscode usage is atypical. @hokein @kbobyrev WDYT about writing .clangd to workspace folders?

I think that depends on everyone's workflow. Just speaking for myself, the current model (assuming the workspacefolders[0] as the project root) at lease works for me, because I use `llvm-project` directory in vscode during the llvm development.

I think this is a useful feature, but the current model seems fragile to me:

- for a monolithic codebase, people working on a subproject likely use the subdirectory (e.g. clang_tools_extra/clangd) as the workspace folder in vscode. For this case, the command will create `.clangd` in the `llvm-project/clang_tools_extra/clangd` directory, but we expect that the project config should be in `llvm-project/`.
- should we respect the currently-editing file in vscode (I think yes)? Think about a case where we have `llvm-project` workspace folder in vscode, then we open and *edit* an out-of-workspace-folder file `/tmp/t.cpp`, now we invoke a `open project config` command,  should we open the .clangd in llvm-project, or there is no project config (respecting the currently-editing `/tmp/t.cpp`), or create a .clangd in `/tmp/`? 
- another observation is that we are mirroring the config behavior in vscode client, this might lead to subtle diverging bugs in the future. 

I was going to suggest that we can add some extensions like `getConfigInfo` in clangd server, then the client can query clangd to get the config information, but realized that clangd doesn't know the concept of project directory :(

src/extension.ts
```diff
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 
 import {ClangdContext} from './clangd-context';
+import * as OpenConfig from './open-config';
```
sam-mccall, line 4: nit: OpenConfig -> openConfig

For whatever reason, our idiom is to features from clangd-context.ts, passing in ClangdContext instead of ExtensionContext.
(I think it might be something to do with allowing the server to be restarted, we unload and reload all features. It's not actually important for this feature but best to be consistent)

sam-mccall (approved):
LGTM with one nit around initialization that I can fix myself.

---

Project config:

This isn't going to free people from having to know where `.clangd` files go, but it's going to put them in a predictable place that will work a good fraction of the time. I think it's a fair tradeoff.

> I was going to suggest that we can add some extensions like getConfigInfo in clangd server, then the client can query clangd to get the config information, but realized that clangd doesn't know the concept of project directory :(

This is obviously pretty heavyweight. I can imagine some refinements to make it work better, but I don't think it's worth it.

---

It'd be nice to expose these from the status bar item, but that doesn't look possible as we have a primary action already (open log) and https://github.com/microsoft/vscode/issues/27196 isn't done.

src/extension.ts
njames93: The reason I passed the ExtensionContext is because the commands don't actually depend on the clangd server being active. However there's very little real world difference in using ClangdContext and consistency is a good thing.

PR #188, opened 2021-05-10
Extend inlay hint support for type hints

HighCommander4:
Corresponding server-side patch: https://reviews.llvm.org/D102148


HighCommander4:
Requesting review, the server-side patch has merged but we need this client-side bit for users to actually see the type hints.

HighCommander4:
Review ping :)

HighCommander4:
@sam-mccall pinging again :) This is a small patch to enable type hints on the client side, so that the server-side type hints patches that we've been merging can actually be used. Would be great to merge!

sam-mccall (approved):
Thanks for the pings, sorry about delaying this

PR #193, opened 2021-05-25
Improve presentation of inactive region highlight

HighCommander4:
This patch intercepts the semantic tokens requests using middleware,
extracts the inactive code ranges, and uses setDecorations() to
provide whole-line background highlighting for them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,9 +157,30 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Warn about conflicting extensions and suggest disabling them."
+                },
+                "clangd.inactiveRegions.useBackgroundHighlight": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Use a background highlight rather than opacity to identify inactive preprocessor regions."
+                },
+                "clangd.inactiveRegions.opacity": {
+                    "type": "number",
+                    "default": 0.55,
+                    "description": "Opacity of inactive regions (used only if clangd.inactiveRegions.useBackgroundHighlight=false)"
                 }
             }
         },
+        "colors": [
+            {
+                "id": "clangd.inactiveRegions.background",
+                "description": "Background color of inactive code regions (used only if clangd.inactiveRegions.useBackgroundHighlight=true)",
+                "defaults": {
+                    "dark": "#1212124C",
+                    "light": "#DCDCDC4C",
+                    "highContrast": "#FCFCFC4C"
+                }
+            }
+        ],
         "commands": [
             {
                 "command": "clangd.switchheadersource",
--- a/src/inactive-regions.ts
+++ b/src/inactive-regions.ts
@@ -1,0 +1,108 @@
+import * as vscode from 'vscode';
+import * as vscodelc from 'vscode-languageclient/node';
+
+import {ClangdContext} from './clangd-context';
+import * as config from './config';
+
+// Parameters for the inactive regions (server-side) push notification.
+interface InactiveRegionsParams {
+  // The text document that has to be decorated with the inactive regions
+  // information.
+  textDocument: vscodelc.VersionedTextDocumentIdentifier;
+  // An array of inactive regions.
+  regions: vscodelc.Range[];
+}
+
+// Language server push notification providing the inactive regions
+// information for a text document.
+export const NotificationType =
+    new vscodelc.NotificationType<InactiveRegionsParams>(
+        'textDocument/inactiveRegions');
+
+export function activate(context: ClangdContext) {
+  const feature = new InactiveRegionsFeature(context);
+  context.client.registerFeature(feature);
+  context.client.onNotification(NotificationType,
+                                feature.handleNotification.bind(feature));
+}
+
+export class InactiveRegionsFeature implements vscodelc.StaticFeature {
+  private decorationType?: vscode.TextEditorDecorationType;
+  private files: Map<string, vscode.Range[]> = new Map();
+  private context: ClangdContext;
+
+  constructor(context: ClangdContext) { this.context = context; }
+
+  fillClientCapabilities(capabilities: vscodelc.ClientCapabilities) {
+    // Extend the ClientCapabilities type and add inactive regions
+    // capability to the object.
+    if (capabilities.textDocument) {
+      const textDocumentCapabilities: vscodelc.TextDocumentClientCapabilities&
+          {inactiveRegionsCapabilities?: {inactiveRegions: boolean}} =
+          capabilities.textDocument;
+      textDocumentCapabilities.inactiveRegionsCapabilities = {
+        inactiveRegions: true,
+      };
+    }
+  }
+  initialize(capabilities: vscodelc.ServerCapabilities,
+             documentSelector: vscodelc.DocumentSelector|undefined) {
+    const serverCapabilities: vscodelc.ServerCapabilities&
+        {inactiveRegionsProvider?: any} = capabilities;
+    if (serverCapabilities.inactiveRegionsProvider) {
+      if (config.get<boolean>('inactiveRegions.useBackgroundHighlight')) {
+        this.decorationType = vscode.window.createTextEditorDecorationType({
+          isWholeLine: true,
+          backgroundColor:
+              new vscode.ThemeColor('clangd.inactiveRegions.background'),
+        });
+      } else {
+        this.decorationType = vscode.window.createTextEditorDecorationType({
+          isWholeLine: true,
+          opacity: config.get<number>('inactiveRegions.opacity').toString()
+        });
+      }
+      this.context.subscriptions.push(
+          vscode.window.onDidChangeVisibleTextEditors(
+              (editors) => editors.forEach(
+                  (e) => this.applyHighlights(e.document.uri.toString()))));
+      this.context.subscriptions.push(
+          vscode.workspace.onDidChangeConfiguration((conf) => {
+            if (!conf.affectsConfiguration('workbench.colorTheme'))
+              return;
+            vscode.window.visibleTextEditors.forEach((e) => {
+              if (!this.decorationType)
+                return;
+              const ranges = this.files.get(e.document.uri.toString());
+              if (!ranges)
+                return;
+              e.setDecorations(this.decorationType, ranges);
+            });
+          }));
+    }
+  }
+
+  handleNotification(params: InactiveRegionsParams) {
+    const fileUri = params.textDocument.uri;
+    const ranges: vscode.Range[] = params.regions.map(
+        (r) => this.context.client.protocol2CodeConverter.asRange(r));
+    this.files.set(fileUri, ranges);
+    this.applyHighlights(fileUri);
+  }
+
+  applyHighlights(fileUri: string) {
+    const ranges = this.files.get(fileUri);
+    if (!ranges)
+      return;
+    vscode.window.visibleTextEditors.forEach((e) => {
+      if (!this.decorationType)
+        return;
+      if (e.document.uri.toString() !== fileUri)
+        return;
+      e.setDecorations(this.decorationType, ranges);
+    });
+  }
+
+  getState(): vscodelc.FeatureState { return {kind: 'static'}; }
+  dispose() {}
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## Next (not yet released)
+
+* Combine inactive region style with client-side (textmate) token colors [#193](https://github.com/clangd/vscode-clangd/pull/193).
+  Requires clangd 17 or later.
+  * The default inactive region style is reduced opacity. The opacity level can be
+    customized with `clangd.inactiveRegions.opacity`.
+  * An alternative inactive region style of a background highlight can be enabled with
+    `clangd.inactiveRegions.useBackgroundHighlight=true`. The highlight color can be
+    customized with `clangd.inactiveRegions.background` in `workbench.colorCustomizations`.
+
 ## Version 0.1.24: April 21, 2023
 
 - Fix an undefined object access in ClangdContext.dispose() [#461](https://github.com/clangd/vscode-clangd/pull/461)
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -5,6 +5,7 @@ import * as ast from './ast';
 import * as config from './config';
 import * as configFileWatcher from './config-file-watcher';
 import * as fileStatus from './file-status';
+import * as inactiveRegions from './inactive-regions';
 import * as inlayHints from './inlay-hints';
 import * as install from './install';
 import * as memoryUsage from './memory-usage';
@@ -161,6 +162,7 @@ export class ClangdContext implements vscode.Disposable {
     memoryUsage.activate(this);
     ast.activate(this);
     openConfig.activate(this);
+    inactiveRegions.activate(this);
     this.client.start();
     console.log('Clang Language Server is now active!');
     fileStatus.activate(this);


HighCommander4:
@sam-mccall I wonder what you think of this patch.

On the one hand, it may seem like a lot of effort for a small cosmetic improvement. (But I've already done the effort :-) )

On the other hand, I believe visualizing inactive code with a background highlight is a common C++ IDE trope that may be worth emulating. This patch is also a good building block for the future enhancement described [here](https://github.com/clangd/clangd/issues/132#issuecomment-864222394) (basically, to omit the semantic foreground color for inactive tokens altogether, such that the **client-side** highlightings (e.g. bold keywords) are retained and combined with the semantic background in inactive code).

The approach taken in this patch to implement the background highlight is in line with [Microsoft's suggestion](https://github.com/microsoft/vscode/issues/123352#issuecomment-856255318) to use editor decorations for this purpose until a more specific API greared towards inactive code is added.

HighCommander4:
@sam-mccall review ping :)

HighCommander4:
Shopping around for some other reviewers :)

Trass3r:
Seems to work.
Still relies on the `comment` semantic token type, which seems to override the default syntax highlighting.
Is it perhaps technically possible to remove the semantic token after processing it so you get a faded region with basic syntax highlighting?

HighCommander4:
> Seems to work. Still relies on the `comment` semantic token type, which seems to override the default syntax highlighting. Is it perhaps technically possible to remove the semantic token after processing it so you get a faded region with basic syntax highlighting?

I had the same thought in https://github.com/clangd/clangd/issues/132#issuecomment-864222394. Haven't tried to see if it's possible yet.

Trass3r:
The thing is that if you don't specify `color` in `createTextEditorDecorationType` nothing gets applied and the code is colored green. I think there was a vscode issue about that.

rolandbernard:
Hi, I just wanted to chime in and say that I implement this in my fork (https://github.com/rolandbernard/vscode-clangd/commit/4f2c9f3f7aa65f129b7e7b424719a7c4f4eec548) because it bugged be so much. I wanted it to look like in the official Microsoft C++ extension, i.e., the region only has a lower opacity and still uses the default syntax highlighting. So I replace all the comment semantic tokens before returning them.

image
(This is how it looks with my changes)

HighCommander4:
> Hi, I just wanted to chime in and say that I implement this in my fork ([rolandbernard@4f2c9f3](https://github.com/rolandbernard/vscode-clangd/commit/4f2c9f3f7aa65f129b7e7b424719a7c4f4eec548))

This is really neat!

It's the same high-level idea that I was after: combine a presentational style indicating "inactive", with client-side (TextMate) token colors. The only difference is the choice of presentation style for "inactive" (in my case it's a grey background, in yours it's lower opacity). Anyways, I believe your patch demonstrates that this sort of combination is possible!

HighCommander4:
@rolandbernard I looked through your patch. One thing I'm curious about is `replaceInactiveRanges()`: is the purpose of changing the token type to `inactiveCodeTokenTypeReplaceIndex` that we know that vscode will not have any style associated with that token type, so it'll just ignore it? And it's simpler to overwrite the token type with a no-op token type, than to erase the token from the encoded data altogether?

HighCommander4:
> Yes, I choose `inactiveCodeTokenTypeReplaceIndex` to be outside the range of token types that are in `capabilities.semanticTokensProvider.legend.tokenTypes`. It seems vscode ignores them, but I have no idea if that is something one can or should rely on. I think it would also be possible to remove the tokens from the data altogether, but this was, as you say, simpler to implement.

Thanks, that makes sense.

The other trick in your implementation that hadn't occurred to me was that the `provideDocumentSemanticTokensEdits` middleware is allowed to pass a full set of tokens (not a delta) on to vscode (I thought I'd have to compute a modified delta that would apply to the modified tokens, which seemed involved).

HighCommander4:
@rolandbernard I updated the PR to make use of the techniques from your patch to preserve the client-side coloring, hope that's cool (I added a comment to credit you).

The styling in my patch currently remains a background highlight, but it's very easy to change to opacity if that's preferred, I don't really feel strongly about it.

Now, we just need someone kind enough to review and consider merging this improvement :)

HighCommander4:
> My conclusion is we should back up and try again:
> 
>    1. remove inactive regions from syntax highlighting as a failed experiment.
> 
>    2. keep the implementation and expose it as a separate LSP extension

Square one, it's nice to see you again :)

This is the very first approach I took for inactive regions (modulo the "combining client-side tokens" part), in Sept 2019:

[Server-side patch](https://reviews.llvm.org/D67536?vs=on&id=220042#toc) (looking at the first version of the diff in particular, as later versions reformulate it on top of semantic highlighting)
[Client-side patch](https://reviews.llvm.org/D67537?vs=on&id=220043#toc) (likewise)

sam-mccall:
Wow, I'd forgotten. That's excruciating to read, I'm sorry.

Still, `s/a mistake/my mistake/`, how do you think we should proceed?

(If you prefer to go ahead with this version, that seems reasonable. I'd love to know whether you think it's not a mistake vs not worth fixing vs too frustrating to touch anymore would be useful!)

HighCommander4:
I agree that conceptually, a separate protocol for inactive regions makes more sense.

The two reasons I see as most convincing are:
 1. Granularity (line vs. token), including the possibility of whole-line styling for inactive regions.
 2. The usefulness of combining the inactive style with token colors. (Maybe, in the distant future when we can get clangd to build an AST for all preprocessor branches, even with _semantic_ token colors.)

So, I think a reasonable way forward would be to resurrect the patches which add a new protocol, rebase them, and tweak them to preserve the client-side TextMate tokens.

I'm happy to do that (probably after finishing up a couple of other in-progress things, so if you'd like to beat me to it, please feel free).

HighCommander4:
>     2. The usefulness of combining the inactive style with token colors. (Maybe, in the distant future when we can get clangd to build an AST for all preprocessor branches, even with _semantic_ token colors.)

Actually, maybe even in the not-so-distant future by sending pseudo-parser based heuristic semantic tokens code inside inactive regions :)

HighCommander4:
I guess, for completeness, it's worth noting that a downside of the dedicated protocol is that we lose the ability of other clients to have _some_ styling of inactive code without implementing the new protocol. (Perhaps we should mitigate that by still sending the comment semantic tokens if the client doesn't indicate capability for the new protocol?)

Trass3r:
> The styling in my patch currently remains a background highlight, but it's very easy to change to opacity if that's preferred, I don't really feel strongly about it.

The opacity really makes the difference in making it clear the code is inactive.
But actually combining it with the background is even better.

```ts
    vscode.window.createTextEditorDecorationType({
      isWholeLine: true,
      opacity: config.get<number>('clangd.inactiveRegions.opacity').toString(),
      backgroundColor:
          new vscode.ThemeColor('clangd.inactiveRegions.background')
    });
```

lua9520:
Any progress about this pr?

HighCommander4:
> Any progress about this pr?@HighCommander4

It's on my list to implement the approach described in [this comment](https://github.com/clangd/vscode-clangd/pull/193#issuecomment-1045292152) but I haven't gotten to it yet. Hope to get to it in the coming weeks, or if someone would like to take it over, please feel free.

thomasthorsen:
Would really like to see this implemented, or at least an option to disable the comment semantic on inactive regions.

We have many build configs for the same source code base and all variants needs to be maintained, but the majority appear as unreadable comments, effectively making it impossible to use this plugin with the current behaviour 😢

I definitely wouldn't mind a pragmatic approach for now, and then a more architecturally clean approach later. I don't know how to apply the patches above to an actual vscode install that is running this extension.

HighCommander4:
Sorry folks, I haven't gotten to this, and I'm not likely to for several more weeks as I'll be travelling.

If someone would like to take this over, you're very welcome to.

If someone would like to pursue a quicker/easier fix, one that could be considered to add an option to the server to not send inactive region tokens altogether (or to the client to ignore them?)

jk000:
@thomasthorsen Here are steps to apply the patch:

Following from https://code.visualstudio.com/api/working-with-extensions/publishing-extension#installation, download/install [Node.js](https://nodejs.org/) then run `npm install -g vsce`. Clone [this repo](https://github.com/clangd/vscode-clangd.git), checkout the latest tag, and apply the patch given by @rolandbernard.  I also had to run `npm install esbuild` and `npm install` before creating the package with `vsce package`.

I had to add the line `getState(): vscodelc.FeatureState { return {kind: 'static'}; }` right above the line with `dispose` in it in `inactive-regions.ts` for it to install correctly due to updates since the patch was created.

Now you can copy the extension that ends in `.vsix` to a remote server and use the `Install from VSIX...` command to install it.

GitMensch:
Hoping for resolved conflicts and possible integration.

miguno:
> Would really like to see this implemented, or at least an option to disable the comment semantic on inactive regions. (https://github.com/clangd/vscode-clangd/pull/193#issuecomment-1195225221)

+1

I have run into the problem of not being able to disable semantic highlighting via clangd for inactive regions when writing `.h` header files that define short-named aliases for long-named functions:

```c
/* myheader.h */

extern int my_super_long_function(int a, int b);

#ifdef MY_SHORT_NAMES
#define short_fn my_super_long_function
#endif
```
Today, the last three lines (ifdef to endif) are always grayed out like comments, which makes maintaining such code sections quite cumbersome:

Screenshot 2023-01-27 at 20 38 28

I *do* like the graying-out of inactive regions when working on `.c` files. It's inconvenient when working on `.h` header files, however.

jay-zhou-yan:
Any progress about this pr? Always looking forward to this official improvement. @HighCommander4

HighCommander4:
Apologies for taking so long to get back to this.

I've pushed an update to the PR which, along with [this server side patch](https://reviews.llvm.org/D143974), implement the approach discussed in [this comment](https://github.com/clangd/vscode-clangd/pull/193#issuecomment-1044315732).

src/inactive-regions.ts
```diff
@@ -0,0 +1,94 @@
...
+  initialize(capabilities: vscodelc.ServerCapabilities,
+    documentSelector: vscodelc.DocumentSelector | undefined) {
+    const serverCapabilities: vscodelc.ServerCapabilities & {
+      inactiveRegionsProvider?: any
+    } = capabilities;
+    if (serverCapabilities.inactiveRegionsProvider) {
+      this.decorationType = vscode.window.createTextEditorDecorationType({
+        isWholeLine: true,
+        light: {
+          backgroundColor: "rgba(220, 220, 220, 0.3)",
```
hokein, line 59: A native question, how do you come up with these numbers? I guess we want to follow the inactive-style of the TS in vscode, but I didn't manage to figure out the color style there (the `Developer: Inspect Editor Tokens and Scopes`
command doesn't seem to give useful information there) .

src/inactive-regions.ts
```diff
@@ -0,0 +1,94 @@
+import * as vscode from 'vscode';
+import * as vscodelc from 'vscode-languageclient/node';
+
+import {ClangdContext} from './clangd-context';
+
+// Parameters for the inactive regions (server-side) push notification.
+// Mirrors the structure in the inactive regions proposal for LSP.
```
hokein, line 7: is there a related GitHub thread for this? If so, it would be nice to mention it (I tried to search `inactive regions proposal for LSP`, it didn't find anything).

hokein (commented):
Thanks for working on this.

src/inactive-regions.ts
HighCommander4: The comment dates back to [this 2019 patch](https://reviews.llvm.org/D67537?vs=on&id=220043#toc). Unfortunately, I no longer remember what it was referring to :/

I do see a similar structure used in both [cquery](https://github.com/jacobdufault/cquery/blob/9b80917cbf7d26b78ec62b409442ecf96f72daf9/src/message_handler.cc#L15-L22) and [vscode-cpptools](https://github.com/microsoft/vscode-cpptools/blob/c417da6ed438327a39b7fa0d18bfd4d05e3279cc/Extension/src/LanguageServer/client.ts#L222), but I can't find an actual LSP proposal based on either of these.

I will remove this comment line to avoid confusion. (And we can make an LSP proposal ourselves once clangd has this support.)

src/inactive-regions.ts
HighCommander4: The colors also date back to the [2019 patch](https://reviews.llvm.org/D67537?vs=on&id=220043#toc). My best guess as to where I got them from is the colors used by Eclipse CDT in this situation.

That said, your comment reminds me that we should probably make these colors customizable via vscode settings.

Another customization that we could allow is to optionally use a reduced opacity of the foreground colors, instead of a background color (some users have expressed a preference for this visual style).

lichjian:
这个问题还没解决吗

HighCommander4:
Updated to address review comments. Notable changes:

 * Background color is now configurable
 * Optionally, opacity can be used instead of a background color (opacity level is also configurable)

HighCommander4:
clang-formatted patch

GitMensch:
Thank you for your work on this. While we wait for the review and pull, would you mind sharing pictures of "old presentation" along with the new options?

HighCommander4:
Fixed a typo in package.json

HighCommander4:
> Thank you for your work on this. While we wait for the review and pull, would you mind sharing pictures of "old presentation" along with how the new options look like?

Sure. Before this patch:

before

After this patch (background highlight option):

after-bg

After this patch (opacity option):

after-opacity

src/inactive-regions.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  handleNotification(params: InactiveRegionsParams) {
+    const fileUri = params.textDocument.uri;
+    const ranges: vscode.Range[] = params.regions.map(
+        (r) => this.context.client.protocol2CodeConverter.asRange(r));
+    this.files.set(fileUri, ranges);
+    this.applyHighlights(fileUri);
+  }
+
+  applyHighlights(fileUri: string) {
```
hokein, line 80: If the user changes their color theme (from `light` => `dark`), we'll not reapply the hightlights, we probably need to add a listener to handle this case, e.g.

```
   
    this.subscriptions.push(
        vscode.workspace.onDidChangeConfiguration((conf) => {
          if (!conf.affectsConfiguration('workbench.colorTheme'))
            return;
           // reapply highlights for inactive region.
        }));
```

src/inactive-regions.ts
hokein: VSCode already provides `editorUnnecessaryCode.border` and `editorUnnecessaryCode.opacity` fields (see https://github.com/microsoft/vscode/blob/main/src/vs/editor/common/core/editorColorRegistry.ts#L44-L45), they are used for unused diagnostics, the rendering style looks very similar to what we propose here (except the background color) and they are theme-aware, can we just use them in our extension to render the inactive regions rather than adding our own ones?  

The downside is that we don't provide many customizations (no foreground&background color), it is fine to not supporting that from my perspective.

hokein (commented):
Thanks, the implementation looks good in general.

I have a question about the rendering theme, see my comment.

src/inactive-regions.ts
HighCommander4: I am open to using these instead of our own opacity setting in the `useOpacity=true` branch. (I would prefer to keep the `useOpacity=false` branch with the background highlight as an option for users who prefer this look.)

However, I'm not quite sure how to make use of this from the extension. I can create a `new vscode.ThemeColor('editorUnnecessaryCode.opacity')`, but what do I do with it?
 * The `ThemeColor` does not expose the actual color or its opacity
 * The `opacity` field of `ThemableDecorationRenderOptions` does not accept a `ThemeColor`.
 * The `color` field of `ThemableDecorationRenderOptions` accepts a `ThemeColor`, but if I pass `new vscode.ThemeColor('editorUnnecessaryCode.opacity')` to it, then it uses this as the foreground color for inactive regions. While the color does have a reduced opacity, it also overwrites the colors from the textmate tokens, thus defeating the purpose of this patch (which is to combine the opacity with the colors of the textmate tokens).

Is there some other way of using this opacity field that I'm overlooking?

src/inactive-regions.ts
HighCommander4: Good catch, added!

HighCommander4:
Addressed review comments, except for the one about `editorUnresolvedOpacity` which I couldn't figure out how to use (details in comment reply)

HighCommander4:
@hokein, perhaps we could proceed with the current patch for now? If at some point we figure out how to make use of vscode's `editorUnnecessaryCode.opacity` instead of having our own setting, we can always switch to that in a follow-up change.

package.json
```diff
@@ -350,6 +360,17 @@
                     "when": "clangd.ast.hasData"
                 }
             ]
-        }
+        },
+        "colors": [
```
hokein, line 364: nit: can we move this section above to the `commands` section? so that all inactiveRegion settings are closer.

src/inactive-regions.ts
hokein: Thanks for exploring it. I thought these fields are exposed and can be accessed via the VSCode public API, looks like they are not.

hokein (approved):
The code looks great (sorry for taking it so long). Let's ship it!

package.json
```diff
@@ -157,6 +157,16 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Warn about conflicting extensions and suggest disabling them."
+                },
+                "clangd.inactiveRegions.useOpacity": {
+                    "type": "boolean",
+                    "default": false,
```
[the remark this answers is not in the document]
GitMensch: That is a change to the current behaviour - but I'd prefer the default to be `true`, too.

In any case the new options (and possible change of default behaviour) should be documented in https://github.com/clangd/vscode-clangd/blob/master/CHANGELOG.md (but I don't know how / when this project changes this file).

HighCommander4:
Rebased and addressed review comments.

package.json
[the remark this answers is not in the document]
HighCommander4: I'm fine with that, I updated the patch to use opacity as the default.

I did invert the naming of the pref (I called it `clangd.inactiveRegions.useBackgroundHighlight` with default value `false`), because it's users who want the non-default (background highlight) behaviour who are more likely to be looking for the pref.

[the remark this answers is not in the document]
HighCommander4: > That is a change to the current behaviour

This PR is a change to the current behaviour either way: the previous behaviour was that inactive regions used a different foreground color, now they use a different opacity or a different background color.

> In any case the new options (and possible change of default behaviour) should be documented in https://github.com/clangd/vscode-clangd/blob/master/CHANGELOG.md (but I don't know how / when this project changes this file).

Good idea, thanks.

I added a "Next" section to the changelog, documenting this change and the configuration options. When the next version is actually released, "Next" can be replaced with the version number and release date.

miguno:
Thanks everyone for completing this!

khlebnikov:
Thank you for making this happen!  I do wonder when is this feature be released?  IIUC, this change was merged in May, and the latest release was in April.  Is it on the horizon or would you recommend just building the extension from source?

khlebnikov:
I don't quite understand. I can use the weekly builds of `clangd` [itself](https://github.com/clangd/clangd/releases), but this is a change to the VS Code extension and I don't see something similar for the [extension](https://github.com/clangd/vscode-clangd/releases).  If I'm looking in the wrong place - please let me know.

HighCommander4:
I sent out a PR to release a new version: https://github.com/clangd/vscode-clangd/pull/516.

I don't believe we currently have any sort of automatically created preview/snapshot releases for vscode-clangd. I agree that this would be nice to have. I'm not sure that "nightly" is the appropriate frequency, given that the project gets commits much less often than that (there have been a total of 6 commits in 2023 so far). Perhaps it would make sense to have every commit trigger a snapshot release? If someone has the familiarity with Github Actions to write a patch to the yaml files in [.github/workflows](https://github.com/clangd/vscode-clangd/tree/master/.github/workflows) that accomplishes this, that would be very neat.

HighCommander4:
> I don't believe we currently have any sort of automatically created preview/snapshot releases for vscode-clangd. I agree that this would be nice to have. I'm not sure that "nightly" is the appropriate frequency, given that the project gets commits much less often than that (there have been a total of 6 commits in 2023 so far). Perhaps it would make sense to have every commit trigger a snapshot release? If someone has the familiarity with Github Actions to write a patch to the yaml files in [.github/workflows](https://github.com/clangd/vscode-clangd/tree/master/.github/workflows) that accomplishes this, that would be very neat.

Filed https://github.com/clangd/vscode-clangd/issues/517 to keep any further discussion of this in a dedicated place.

kon72:
Hello @HighCommander4,

The highlighting appears to be ineffective in certain instances, especially on Windows.
I've sent out a fix in #515. Can you please review it?

menash134:
@HighCommander4  Thanks for the cool work. Could somebody point out the documentation for highlighting the disabled code. Iam using Nvim with clangd, which option needs to be enabled

HighCommander4:
> @HighCommander4 Thanks for the cool work. Could somebody point out the documentation for highlighting the disabled code. Iam using Nvim with clangd, which option needs to be enabled

I've implemented the server-side of the inactive region highlighting in clangd, and the client-side support in vscode.

Client-side support in other clients like neovim needs to be implemented separately in the respective client. I don't know whether anyone has done that for neovim yet.

PR #209, opened 2021-07-09
Depend on new version of auto-installer

kadircet:


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-clangd",
     "displayName": "clangd",
     "description": "C and C++ completion, navigation, and insights",
-    "version": "0.1.11",
+    "version": "0.1.12",
```
hokein, line 5: This bump will cause a new release of vscode-clangd extension automatically. I think we should do that in a separated PR.

kadircet: thanks for letting me know, done!

Issue #162, opened 2021-03-05
Add an option to stop showing warning about clangd + mscpp incompatibility


kirillbobyrev:
Context: https://github.com/clangd/vscode-clangd/pull/141#issuecomment-784561701

sam-mccall:
Maybe this shouldn't be a visible option, just a way to "never ask again"

kirillbobyrev:
Yes, in my mind it looks like a button next to "Disable mscpp" on the
warning.

theswordsmahin:
I agree with the above here. Our project was setting 
```
    "C_Cpp.autocomplete": "Disabled",  // So you don't get autocomplete from both extensions.
    "C_Cpp.errorSquiggles": "Disabled", // So you don't get error squiggles from both extensions
```
As clang's errors/autocomplete were much better, but leaving the `intellisenseEngine` enabled for the enhanced colorization

So an option to disable the warning would be great.

matms:
I'm not very familiar with vscode-clangd's contribution process, but I also faced the same issue, so I have made a small patch adding a config option to disable this warning (https://github.com/matms/vscode-clangd/commit/95e90f77bc447f06959174b834dc06657ab34680).

I'm not sure if I should make a pull request?

HighCommander4:
> I'm not sure if I should make a pull request?

I would say yes, go ahead

kirillbobyrev:
> I'm not very familiar with vscode-clangd's contribution process, but I also faced the same issue, so I have made a small patch adding a config option to disable this warning ([matms@95e90f7](https://github.com/matms/vscode-clangd/commit/95e90f77bc447f06959174b834dc06657ab34680)).
> 
> I'm not sure if I should make a pull request?

Absolutely!

Sorry, this is actually my bad for forgetting about this and not delivering on my promise to fix this originally. I'll gladly review your patch and provide help if needed.

kirillbobyrev:
Seems to be fixed, makes sense to close.

kirillbobyrev:
The new GitHub UI makes me miss the button every time 😆 Thanks, @HighCommander4!

PR #221, opened 2021-08-16
Add option to suppress warning about the Microsoft C/C++ extension incompatibility

matms:
I've added an option to control suppression of the incompatibility warning (and not change the C/C++ settings).

I'm making this pull request as suggested in https://github.com/clangd/vscode-clangd/issues/162, since others seem to have faced this same issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,6 +148,11 @@
                         "Automatically restart the server",
                         "Do nothing"
                     ]
+                },
+                "clangd.detectAndDisableExtensionConflicts": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Whether to look for extension conflicts and disable any conflicting settings."
                 }
             }
         },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -148,6 +148,11 @@
                         "Automatically restart the server",
                         "Do nothing"
                     ]
+                },
+                "clangd.detectAndDisableExtensionConflicts": {
```
kirillbobyrev, line 152: nit: how about `clangd.detectExtensionConflicts` + `Warn about conflicting extensions and suggest disabling them.`?

src/extension.ts
```diff
@@ -27,23 +27,28 @@ export async function activate(context: vscode.ExtensionContext) {
...
+      if (cppTools && cppTools.isActive) {
+        const cppToolsConfiguration = vscode.workspace.getConfiguration('C_Cpp');
+        const cppToolsEnabled = cppToolsConfiguration.get('intelliSenseEngine');
+        if (cppToolsEnabled !== 'Disabled') {
+          vscode.window
+              .showWarningMessage(
+                  'You have both the Microsoft C++ (cpptools) extension and ' +
+                      'clangd extension enabled. The Microsoft IntelliSense features ' +
+                      'conflict with clangd\'s code completion, diagnostics etc.',
+                  'Disable IntelliSense')
```
kirillbobyrev, line 45: Nit: I think the perfect UX/UI would be adding a button `Don't show this warning`.

Something like 

```typescript
vscode.window.showWarningMessage('', ['Disable IntelliSense', 'Don\'t show this warning']).
    then(selection => {
        if (selection = 'Don\'t show this warning') {
            /* update clangd.detectAndDisableExtensionConflicts similar to what we
               have with C_Cpp.intelliSenseEngine right now */
        }
    }
```

kirillbobyrev (changes requested):
Hey, sorry, I originally missed the patch. Thanks for picking it up and working on this! The logic looks good, I have a couple of suggestions that would make this patch even more awesome, would be happy to accept once resolved!

kirillbobyrev:
(also, please `clang-format` the code before submitting, otherwise the CI linters will complain)

matms:
Done. This patch also means you don't have to restart VsCode, since it should now clear the interval upon pressing 'Never show this message'.

Note that if this option is later used for _other_ incompatibilities, then pressing 'Never show this message' would disable all of them.

kirillbobyrev:
Tested, works great, thank you for working on this!

PR #222, opened 2021-08-18
Update README.md

playgithub:
update ways to referecne `compile_commands.json`


README.md
```diff
@@ -30,7 +30,18 @@ can usually be generated by your build system
 or with
 [many other tools](https://sarcasm.github.io/notes/dev/compilation-database.html)).
 
-It should live at the top of your source tree: symlink or copy it there.
+Different ways to reference `compile_commands.json`:
```
kirillbobyrev, line 33: How about

```markdown
You can make `compile_commands.json` discoverable by:

* Copying or making a symlink for the file to the root directory of your source tree
* Through [clangd configuration mechanism](https://clangd.llvm.org/config.html)
<YAML BLOCK>
If:
  # Note: This is a regexp, notice '.*' at the end of PathMatch string.
  PathMatch: /path/to/project/.*
CompileFlags:                      
  CompilationDatabase: /path/to/compile_commands/dir
</YAML BLOCK>
```

For the CLI flags, this is what it'd look like.

```markdown
* Through clangd command line arguments and VSCode settings (not recommended): put the following into `/path/to/project/.vscode/settings.json` (local) or global configuration
<JSON BLOCK>
{
  "clangd.arguments": [
    "--compile-commands-dir=<path_to_directory_containing_compile_commands.json>"
  ]
}
</JSON BLOCK>
```

However, I think advertising CLI flags usage is probably not what we want, WDYT @sam-mccall?

kirillbobyrev (changes requested):
Hey, thanks for picking it up! I do think that the topic of discovering `compile_commands.json` is quite important and it'd be great to have more visibility.

@sam-mccall @hokein do you think it's also worth highlighting in our "tool" docs ([clangd.llvm.org](https://clangd.llvm.org/))?

README.md
playgithub: Compared to the other two methods, CLI flag doesn't introduce anything in the project itself.
And put config for `clangd` in the configure of the project (`.vscode` in the case) is consistent.

HighCommander4: To me, adding something to `.clangd` doesn't seem any more "in the project itself" than adding something to `.vscode/settings.json`.

Adding something to `.clangd` seems more general because it applies to any client used with clangd.

playgithub: `clangd` is an excellent `language server`, anyway it is for convenience, not part of a project logically.
For example, the dev env is clangd ready, but someone is to use an alternative `language server` instead of `clangd`,
when `.clangd` there, it means `clangd` works, and it wastes cpu adn disk spaces, right?

HighCommander4: The presence of a `.clangd` file is not what triggers clangd to run, if that's what you mean. It's just a file that clangd reads for configuration info.

Starting clangd is done by the client; for vscode, it happens if tht vscode-clangd extension is installed *and* a file with C/C++ flle type is opened.

playgithub: So in vscode 3 requirements below can trigger `clangd` to run, right?
* `vscode-clangd` ready
* `.clangd` ready
* C/C++ file opened

If true, `.clangd` is more like a trigger.

HighCommander4: No, clangd will run even if the `.clangd` file doesn't exist. It's just an optional file where it looks for confinguration info.

playgithub: ok, `.clangd` is an official way, and cli argument has its own benifits.

It's only about preference.

Thanks

playgithub: For there are other assistant files like:
* .git
* .gitingore
* README.md
* ...
It seems more consistent to use .clangd, and it's ready to contain more settings.
However to make a project source tree absolutely clean, a root folder for the project is possible. :nerd_face:

kirillbobyrev: I am not sure I understand the last comment correctly, but also if one wants to keep the source tree free of additional files (though `.gitignore` is really handy for effectively ignoring it through VCS), the original YAML code I posted was meant for the **global** clangd config (living in `~/.config/clangd/config.yaml`). Hence, the `.clangd` is not even needed inside the project folder if one puts `If: PathMatch: /path/to/project/.*` inside the global config next to the modifications one wants to make. The `.clangd` file within source tree would not need to contain these two lines.

playgithub: For comment, a clean source tree contains only files for the project itself without external tools that make developers much easier, like:
* .git
* .gitignore
* .clangd

Because the tools might have alternatives, so tight-bind to the project is not convenient when a developer is going to use an alternative tool. It is unsually to replace `git`, and `clangd` is also common, but it should keep the flexibility for alternatives for less popular tools, and the best way is too keep the source tree clean.

For `~/.config/clangd/config.yaml`, I think it must known where `compile_commands.json` is, this relates to `CMAKE_BINARY_DIR` which tends to change.

sam-mccall:
This is a reasonable idea, but I don't think we should add more detail to the extension README, rather to our main docs.
I've added a README->docs link, and we should work out what needs to be added to https://clangd.llvm.org/installation#project-setup

PR #227, opened 2021-09-07
Allow enabling parameter hints and type hints separately

HighCommander4:


HighCommander4:
This addresses the user request made [here](https://github.com/clangd/clangd/issues/824#issuecomment-884053677)

src/inlay-hints.ts
```diff
@@ -212,15 +213,21 @@ class InlayHintsFeature implements vscodelc.StaticFeature {
...
     const decorations: InlayDecorations = {parameterHints: [], typeHints: []};
     const conv = this.context.client.protocol2CodeConverter;
+    const showParameterHints = config.get<boolean>('inlayHints.parameterHints');
+    const showTypeHints = config.get<boolean>('inlayHints.typeHints');
     for (const hint of hints) {
       switch (hint.kind) {
       case InlayHintKind.Parameter: {
-        decorations.parameterHints.push(
-            parameterHintStyle.toDecoration(hint, conv));
+        if (showParameterHints) {
```
kadircet, line 221: nit: rather than branching for every hint, we can just clear parameterHints/typeHints before returning decorations at the bottom.

kadircet:
Thanks! Having custom UI elements to configure such knobs is definitely great. But I wonder if this will create confusion combined with the fact that user actually needs to provide `--inlay-hints` command line argument && have a recent clangd, in addition to playing with these knobs. I think we should default these options to false, and add `--inlay-hints` to clangd arguments at startup if any of them is true (or by default? since they'll be filtered out). But the problem is it's still unclear whether clangd supports inlay-hints or not. I wonder if it is possible to alter settings dynamically after server capabilities has been published (or some other version check we can perform on the clangd executable).

(As a follow-up) I think this is a configuration that can benefit multiple editors (in the future) and might even use some file-level granularity. So I think we should actually have a server-side configuration option, that can drive the visitors behaviour later on. It would be great if we had support for configuration over LSP, then client-side implementations would look a lot more uniform. Hopefully that's a state we'll get eventually!

HighCommander4:
> But the problem is it's still unclear whether clangd supports inlay-hints or not. I wonder if it is possible to alter settings dynamically after server capabilities has been published (or some other version check we can perform on the clangd executable).

I assume you mean so that we can "hide" these config options if clangd does not support inlay hints?

I haven't been able to find a way to do this in vscode.

> (As a follow-up) I think this is a configuration that can benefit multiple editors (in the future) and might even use some file-level granularity. So I think we should actually have a server-side configuration option, that can drive the visitors behaviour later on. It would be great if we had support for configuration over LSP, then client-side implementations would look a lot more uniform. Hopefully that's a state we'll get eventually!

The LSP does have support for configuration-over-LSP, in the form of the [workspace/configuration](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_configuration) request. I had a look at the vscode client implementation, and it looks like it's wired up such that clangd can just send a request with `{ scopeUri: <fileBeingEdited>, section: "clangd.inlayHints.parameterHints"}`, and vscode will return the value of the `clangd.inlayHints.parameterHints` config option.

However, I'm not clear on how this would co-exist with clangd.yaml. Would the idea be to allow users to specify the settings **either** in clangd.yaml or in the vscode config? Which one should take precedence?

kadircet:
> I assume you mean so that we can "hide" these config options if clangd does not support inlay hints?

I was actually suggesting to not display the option until we are sure clangd is capable. But that's not really the most important problem here. I think we should move with the assumption that `--inlay-hints` will become the default eventually and most users will have a recent enough clangd.

> However, I'm not clear on how this would co-exist with clangd.yaml. Would the idea be to allow users to specify the settings either in clangd.yaml or in the vscode config? Which one should take precedence?

These are all design decisions to be made. But the idea is to enable LSP client to be a different config source (similar to the way we stack up multiple config files on the ancestor paths + user config). That way the version skew between the vscode-clangd extension and clangd binary would be less of an issue, as unknown keys in the configuration can be handled gracefully at runtime.

Please note that I am not suggesting to block this configuration on existence of an LSP-based config provider. I am just saying that we should probably have the knob on the server first, and wait with the settings in the vscode UI (i.e. this patch) until `--inlay-hints` is at least the default in latest release (otherwise, especially in its current state, users will be confused about the fact that inlay hints are enabled in the settings UI, but not shown in the editor). Having an LSP-based config provider just gets rid of the extra waiting step in a principled way for the options we'll introduce in the future.

HighCommander4:
@kadircet Thanks for the additional thoughts. To summarize, I think I've understood two different suggestions for the short term:

 1. Add the options to vscode, defaulting to false, and if the user turns either of them on, then (re-)start clangd with `--inlay-hints`.

 2. Add equivalent options to clangd.yaml instead, allowing the user to control the feature that way.

Which one you would like to proceed with for now? I'm happy to do either.

kadircet:
Yes these were the suggestions but I forgot to mention a big issue with (brought up by Sam):
> Add the options to vscode, defaulting to false, and if the user turns either of them on, then (re-)start clangd with --inlay-hints.

Once extension adds that argument and clangd doesn't know about `--inlay-hints`, consequences are a lot more catastrophic (clangd would crash loop).

Hence let's just go with the second option for now and figure out what to do with configuration through the vscode UI later on (I think we can't really have a useful enough UX for such things until LSP based configuration provider is in-place, due to the lack of version-locking between clangd and extension).

HighCommander4:
@kadircet Regarding the server-side configuration, we'll need to decide how the config options interact with the existing `--inlay-hints` command-line flag.

I imagine two boolean config options `ShowParameterHints` and `ShowTypeHints`, in an `InlayHints` section. I think the interaction with `--inlay-hints` that would make sense is: if `--inlay-hints` is provided, we act as of `ShowParameterHints=true` and `ShowTypeHints=true` were specified (via `FlagsConfigProvider`). However, the user could also choose to specify one or both of these flags without `--inlay-hints`.

I ran into one small wrinkle when implementing this. The way `--inlay-hints` works right now is, it governs whether the server advertises the `clangdInlayHintsProvider` capability, which governs whether the client sends clangd inlay hint requests to begin with.

With the config approach, since the two hint types can be enabled independently, and also because config options can be scoped to a subdirectory, we need to handle the options when responding to requests instead.

So, should we advertise `clangdInlayHintsProvider` unconditionally? Or should we attempt to check something like, "at least one subdirectory has at least one of the options enabled"?

kadircet:
I think we should just assume `--inlay-hints` doesn't exist at all, and we're leaving in a world clangd provides inlay hints by default.

It means `ShowParameterHints` and `ShowTypeHints` will be true by default, and the user will need to explicitly turn them off for parts of the projects they don't want.

In the temporary state, `--inlay-hints` will be guarding everything at a higher level. No matter what we have in the config fragments, if the user doesn't have `--inlay-hints`, clangd won't do anything. I think not displaying inlay hints to the user, even after mentioning them in config fragments, won't be as surprising. As the user needs to learn about this config option first through some documentation (rather than just seeing some buttons in a UI), and we should surely mention that they'll need to pass `--inlay-hints` to clangd, as this is not in the final state yet.

WDYT?

HighCommander4:
Closing in favour of https://reviews.llvm.org/D116713 which implemented the server-side approach (thank you Sam!)

PR #245, opened 2021-11-01
Add switch header/source to editor context menu

### LINKED ISSUE #244: Editor context menu has the wrong "Switch Header/Source" item
After installing the vscode-clangd plugin and disabling `C_Cpp.intelliSenseEngine`, if I use `Alt+O`, I can switch b/w header and source files.

However, on the editor right-click (context) menu, the "Switch Header/Source" item is still bound to the function from `C_Cpp.intelliSenseEngine`, thus flagging an error that intellisense is disabled.

I tried locally adding a menu item in `package.json` of the vscode-clangd extension (just to see if that would work - not in the attached screenshot). That creates a new menu item for this operation and it works fine, but I can't figure out how to remove the menu entry that's already there and is incorrect.

**Logs**
switch-source-header

**System information**
Clangd version (from the log, or `clangd --version`): 13
clangd extension version: 0.1.13
Operating system: Linux / Ubuntu 20.04.

Related: 
- https://github.com/clangd/vscode-clangd/issues/182
- https://github.com/cquery-project/vscode-cquery/issues/48#issuecomment-414185582

vaivaswatha:
Fixes #244.

Note that this displays "clangd: switch between source/header" in the context menu. The "clangd" prefix doesn't look good, but I can't seem to fix that without changing the title of the command (which also affects what one sees in the command palette). I'm not sure what's the best option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -254,6 +254,7 @@
                 },
                 {
                     "command": "clangd.switchheadersource",
+                    "when": "resourceLangId == cpp",
                     "group": "0_navigation@5"
                 },
                 {
--- a/package.json
+++ b/package.json
@@ -252,10 +252,10 @@
                     "group": "0_navigation@4",
                     "_comment": "see https://github.com/microsoft/vscode-references-view/blob/f63eaed9934ca5ecc8f3fb3ca096f38c6e5e181f/package.json#L162"
                 },
-		{
-		    "command": "clangd.switchheadersource",
-		    "group": "0_navigation@5"
-		},
+                {
+                    "command": "clangd.switchheadersource",
+                    "group": "0_navigation@5"
+                },
                 {
                     "command": "clangd.ast",
                     "when": "clangd.ast.supported"
--- a/package.json
+++ b/package.json
@@ -252,6 +252,10 @@
                     "group": "0_navigation@4",
                     "_comment": "see https://github.com/microsoft/vscode-references-view/blob/f63eaed9934ca5ecc8f3fb3ca096f38c6e5e181f/package.json#L162"
                 },
+		{
+		    "command": "clangd.switchheadersource",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -252,6 +252,10 @@
                     "group": "0_navigation@4",
                     "_comment": "see https://github.com/microsoft/vscode-references-view/blob/f63eaed9934ca5ecc8f3fb3ca096f38c6e5e181f/package.json#L162"
                 },
+                {
+                    "command": "clangd.switchheadersource",
+                    "group": "0_navigation@5"
+                },
```
HighCommander4, line 258: We should add `"when": "resourceLangId == cpp"`, to avoid this context menu item showing up in non-C++ files.

(Bonus: `"clang.ast"` should really have this condition to. If you'd like, feel free to submit a separate PR adding that.)

HighCommander4:
Thanks!

> Note that this displays "clangd: switch between source/header" in the context menu. The "clangd" prefix doesn't look good, but I can't seem to fix that without changing the title of the command (which also affects what one sees in the command palette). I'm not sure what's the best option.

I think this is fine for now. Perhaps, in the future, we can consider renaming the command to omit the "clangd: " prefix, similar to how the `clangd.typeHierarchy` and `clangd.AST` commands do not have this prefix.

HighCommander4 (changes requested):

package.json
vaivaswatha: Thank you @HighCommander4. I've added condition now to limit this to C++ files.

HighCommander4 (approved):
Thanks!

HighCommander4:
@kadircet heads up in case you have any thoughts on this, otherwise I think it's ready to merge

kadircet:
thanks!

PR #246, opened 2021-11-09
Add activation events

liorsagy:
When trying to run the `restart language server` and other commands before clangd extension activates, an unhelpful error message stating the command does not exist pops up.


package.json
```diff
@@ -30,7 +30,11 @@
         "onLanguage:objective-cpp",
         "onCommand:clangd.activate",
         "onCommand:clangd.install",
-        "onCommand:clangd.update"
+        "onCommand:clangd.update",
+        "onCommand:clangd.restart",
+        "onCommand:clangd.switchheadersource",
```
kadircet, line 35: `clangd.switchheadersource` and `clangd.restart` won't be really useful if clangd isn't available already. but i suppose it makes sense for `clangd.{project,user}config`.

looping in @hokein as he uses vscode and might have opinions about the matter.

sam-mccall:
What commands do you want to use before clangd is active, and what do you expect them to do?

Issue #247, opened 2021-11-12
"show call hierarchy" gives error "MISSING provider"


sheldonneuberger-sc:
"show call hierarchy" on a function in an objc files doesn't work, and it shows a vscode notification that says "MISSING provider". This was working a couple weeks ago. "Find All References" on the same function works correctly.

**Logs**
Please attach the clangd log from the "Output" window if you can.
If possible, run with `--log=verbose` - note that the logs will include the contents of open files!

clangd output:
```
I[10:12:55.274] <-- textDocument/codeAction(82)
I[10:12:55.275] --> reply:textDocument/codeAction(82) 1 ms
I[10:12:55.275] --> textDocument/clangd.fileStatus
I[10:12:56.229] <-- textDocument/hover(83)
I[10:12:56.230] --> reply:textDocument/hover(83) 1 ms
I[10:12:56.230] --> textDocument/clangd.fileStatus
I[10:12:59.562] <-- textDocument/prepareCallHierarchy(84)
I[10:12:59.563] --> reply:textDocument/prepareCallHierarchy(84) 0 ms
I[10:12:59.563] --> textDocument/clangd.fileStatus
```

**System information**
Clangd version (from the log, or `clangd --version`): 13.0.0
clangd extension version: v0.1.13
Operating system: osx 11.6
vscode: 1.62.2

HighCommander4:
I'm not able to reproduce this.

Could you please provide an example objective-c++ source file that demonstrates the issue, and also verbose logs?

sheldonneuberger-sc:
It's possible it's an issue on my machine then. Will check this week if it repros for a coworker, and I'll try to get some more details.

Is there a way to tell where this "MISSING provider" notification from vscode is coming from? I can't tell if it's emitted from vscode itself or clangd or the clangd extension.

HighCommander4:
> Is there a way to tell where this "MISSING provider" notification from vscode is coming from? I can't tell if it's emitted from vscode itself or clangd or the clangd extension.

I've never seen this error before, and I can't find it in the source code of any of the above.

sheldonneuberger-sc:
This emits at the same time in exthost.log:
```
[2021-11-15 11:44:48.038] [exthost] [error] TypeError: Cannot read property 'kind' of null
    at be (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:94:7738)
    at Array.map (<anonymous>)
[...]

PR #253, opened 2021-11-19
show better error message when no function found during prepareCallHierarchy

sheldonneuberger-sc:
Slight improvement on https://github.com/clangd/vscode-clangd/issues/247, here we now show a "No function found at cursor location." information message if the user selects "Show Call Hierarchy" on a non-function.

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -27,6 +27,25 @@ export function isClangdDocument(document: vscode.TextDocument) {
 }
 
 class ClangdLanguageClient extends vscodelc.LanguageClient {
+
+  // Due to problems like https://github.com/microsoft/TypeScript/issues/30071
+  // and because there's mutiple sendRequest declarations with three params,
+  // we use 'any' types for these three params.
+  async sendRequest(type: any, params: any, token?: any): Promise<any> {
+    const result = await super.sendRequest(type, params, token);
+    // prepareCallHierarchy will always return at least 1 result if it's called
+    // on a valid function/method (because the first entry is the function on
+    // which you called it). If it's called on a non-function we intercept here
+    // to give the user a better error message.
+    if (type?.method === 'textDocument/prepareCallHierarchy') {
+      if (Array.isArray(result) && result.length == 0) {
+        vscode.window.showInformationMessage(
+            'No function found at cursor location.');
+      }
+    }
+    return result;
+  }
+
   // Override the default implementation for failed requests. The default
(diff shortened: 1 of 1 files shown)


kadircet:
We try to keep the vscode-clangd as thin as possible. I am not sure if the extra error message carries enough weight to justify the costs. Why do you think the new UX in vscode with `No results found` is not enough?

sheldonneuberger-sc:
Here's my line of thinking: most users of this extension who see "No results found" will think "clangd doesn't know of any callers for this function". But that's actually not true, what really happened is "clangd doesn't believe this is a function".

So when someone does "Show Call Hierarchy" on something that isn't supported yet by clangd, this gives a clearer picture of what's going on to the user.

I see what you're saying about keeping this extension as thin as possible, although the additional weight here is small. Let's see what @jrieken comes up with on the vscode bug https://github.com/microsoft/vscode/issues/137415, maybe that will avoid this problem entirely.

Thanks for the guidance on this @kadircet.

jrieken:
This is an issue in VS Code and your call hierarchy provider should not be showing/driving the UI. Always consider that other extensions, like LiveShare, can "drive" your provider in different scenarios and that UI messages look very out of place than

kadircet:
> Here's my line of thinking: most users of this extension who see "No results found" will think "clangd doesn't know of any callers for this function". But that's actually not true, what really happened is "clangd doesn't believe this is a function".
>
> So when someone does "Show Call Hierarchy" on something that isn't supported yet by clangd, this gives a clearer picture of what's going on to the user.

I don't feel like there's a big difference between the two. You're right that `No results` is not enough to disambiguate between whether the language server failed to find a `callable` at cursor or failed to find any `callers/callees` for the `callable`. But this won't prevent a user from going and filing a bug if they think there should've been an answer to their request and later on it's on the language server maintainer to try and figure out what went wrong anyways.

> I see what you're saying about keeping this extension as thin as possible, although the additional weight here is small.

I don't agree with this one. Yes it's small in number of lines, but it is actually a whole new concept about intercepting all the calls between VSCode and the languageserver, which is a great power. Hence will require more responsibility in the future to make sure we're not introducing too much new hooks to make the interaction between the client and server surprising. This is a lot of mental effort for maintenance and as @jrieken pointed out it'll also be surprising for users that work with multiple language servers or different plugins.

sheldonneuberger-sc:
Thanks @kadircet and @jrieken for taking the time to explain, it makes sense to me now. Will close this one.

PR #254, opened 2021-11-19
enable switchheadersource for all langs, and add category to all commands

sheldonneuberger-sc:
Based on `activationEvents` in package.json, the plugin is only enabled for c/cpp/cuda/objc/obcpp, and all of those have header/source files (except maybe cuda?). I've confirmed with an objc .m/.h pair that this works correctly.

I also added a category for all commands and removed the clangd prefix from the title. This keeps the "clangd: " prefix on all commands in the command palette, but uses only the title in the context menu (which looks nicer).

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -254,7 +254,6 @@
                 },
                 {
                     "command": "clangd.switchheadersource",
-                    "when": "resourceLangId == cpp",
                     "group": "0_navigation@5"
                 },
                 {
--- a/package.json
+++ b/package.json
@@ -162,7 +162,7 @@
             {
                 "command": "clangd.switchheadersource",
                 "category": "clangd",
-                "title": "Switch Source/Header"
+                "title": "Switch Between Source/Header"
             },
             {
                 "command": "clangd.install",
--- a/package.json
+++ b/package.json
@@ -161,73 +161,88 @@
         "commands": [
             {
                 "command": "clangd.switchheadersource",
-                "title": "clangd: Switch between source/header"
+                "category": "clangd",
+                "title": "Switch Source/Header"
             },
             {
                 "command": "clangd.install",
-                "title": "clangd: Download language server"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -254,7 +269,6 @@
                 },
                 {
                     "command": "clangd.switchheadersource",
-                    "when": "resourceLangId == cpp",
```
kadircet, line 257: indeed there's nothing limiting this feature to cpp files, in theory all clangd functionality is available for all c-family languages supported by clang.

the problem here is though, if we don't have any when/enablement clause it'll show up on right click on every file and it might become annoying if every plugin did that. So can you just extend the condition to cover c/c++/cuda/objc/objc++ according to https://code.visualstudio.com/docs/languages/identifiers#_known-language-identifiers ?

kadircet (changes requested):
thanks, this LG with a small comment about the when clause for right click menu.

package.json
sheldonneuberger-sc: Got it, didn't realize that the context menu appears in all files regardless of the langs listed in `activationEvents`. I changed this to check for the specific c-based languages, and confirmed it's working as expected (shows for objc, doesn't show for py).

package.json
```diff
@@ -161,73 +161,88 @@
         "commands": [
             {
                 "command": "clangd.switchheadersource",
-                "title": "clangd: Switch between source/header"
+                "category": "clangd",
+                "title": "Switch Source/Header"
```
kadircet, line 165: can you preserve the `between` here?

package.json
kadircet: thanks!

kadircet (approved):

package.json
sheldonneuberger-sc: Missed this comment earlier. Fixed.

PR #259, opened 2021-11-29
Move to vscode builtin cuda-cpp language

cloudhan:
Bulitin cuda language was added in https://github.com/microsoft/vscode/commit/95420b39f28dab93e0764d442e74a7984a78ab38

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "activationEvents": [
         "onLanguage:c",
         "onLanguage:cpp",
-        "onLanguage:cuda",
+        "onLanguage:cuda-cpp",
         "onLanguage:objective-c",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "activationEvents": [
         "onLanguage:c",
         "onLanguage:cpp",
-        "onLanguage:cuda",
+        "onLanguage:cuda-cpp",
```
hokein, line 28: This feature is available in vscode 1.56.0 or later. We need to bump the required vscode version (on Line 11).

cloudhan: Done, I also updated the package @types/vscode package to match the this version.

hokein (approved):
Thanks!

PR #261, opened 2021-11-30
export hover api

93Alliance:
I developed a plugin that required clangd's hover api, so i exported the hover api to allow more people to extend it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -337,24 +337,24 @@
             ]
         },
         "colors": [
-          {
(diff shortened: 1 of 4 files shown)


sam-mccall:
Maybe you can elaborate a bit more on what this is for?

I can't understand what you can usefully do with this text other than display it. It's not safe to parse, the contents are not guaranteed between server versions.

Issue #262, opened 2021-12-01
The is no (configurable) delay while typing.


strpbrk:
Allow me to ask for this feature, even though I fear that this feature would need to be implemented in the language client project. Nevertheless the configuration option would need to be supported here...

There doesn't seem to be a configurable, or at least **significant** delay while typing. I would expect that semantic information, or at the very least the list of found **problems** would not jump around while I'm typing. If the former is necessary, at least the latter should not be. The subjective feeling is that I'm working on an old computer with unnecessary distractions jumping all over the screen. (I have auto suggestions and all auto stuff turned off).

If there is nothing you can do, could you please point me to other resources, such as open issues on the language client project, a missing feature elsewhere, or at least an explanation why this would not be possible any time soon. I have tried, but my search has not found anything of significance.

Thanks!

sam-mccall:
The server is in charge of publishing diagnostics. The client could suppress updates to diagnostics while typing - this seems likely to be a vscode-level feature rather than even the vscode-languageclient library. I don't have any particular insight there.

But the server can also delay computing/pushing diagnostics while it's seeing updates from the client.
We actually do have this code in clangd (a debounce delay before reparsing for diagnostics on the server side), but it's tuned for responsiveness: the delay is equal to the parse time for the file, but never less than 50 or more than 500ms.

Before exposing this it would probably be nice to know whether it actually does what you want: these things are about "feel" and hard to predict. 
If you're able to build clangd locally, you could add this line to `clang-tools-extra/clangd/tool/ClangdMain.cpp`, near where other options are set:
```
Opts.UpdateDebounce = DebouncePolicy::fixed(std::chrono::seconds(5));
```

strpbrk:
I have built and tested clangd. Verified that the option is correctly used in "TUScheduler.cpp" to create a Deadline.
This had no effect on the output, certainly no delay of 5s.

However, even if this worked, I would have to disagree that it is a correct solution.
The server should not be concerned about things that alter the user experience. It simply provides a service, and it is just fine that it is tuned for responsiveness.
It is up to the client to do one of two things (or both):
1. Send text changes after a short delay, or after the user stops typing.
2. Apply updates from the server with a delay, or after the user stops typing.
Microsoft's plugin has "intelliSenseUpdateDelay" option to control one of these (although I didn't try their plugin).

[...]

PR #263, opened 2021-12-04
Add basic support for delaying and combining text changes (#262)

strpbrk:
I have managed to implement it with minimal impact on dependencies, and should work correctly up to the latest dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
         "@clangd/install": "0.1.4",
         "abort-controller": "^3.0.0",
         "jsonc-parser": "^2.1.0",
-        "vscode-languageclient": "7.1.0-next.1",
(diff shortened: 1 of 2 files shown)


HighCommander4:
@strpbrk just to confirm my understanding, this patch has the issues described in [this comment](https://github.com/clangd/vscode-clangd/issues/262#issuecomment-1308565341), and is thus not suitable to be reviewed and merged in its current form, right?

HighCommander4:
> @HighCommander4 it has at least those issues. More issues may have been introduced with API changes over the years, which I have not tracked. If I ever make time to get back to this problem, I will make a new evaluation, bug reports, pulls, etc.

Thanks. I've marked this PR as a "draft" to make it clear that it's not ready for review.

PR #273, opened 2022-01-07
Remove legacy semantic-highlighting client code.

hokein:
The legacy semantic-highlighting support has been dropped since clangd13.


sam-mccall:
Wow, I thought this was gone!

Another relevant milestone is clangd 11, which is the first version that supports semanticTokens. (I think it's no longer possible to install new versions of the vscode-clangd on vscode versions so old they don't support semantic tokens).

LG, but I think we should maybe coordinate landing this with the llvm 14 release?
I think it makes it a little easier to understand the schedule: replacement available for 4 releases, server support gone for 2 releases.

PR #276, opened 2022-01-08
Inlay hints: support position, simplify.

sam-mccall:
Previously we had hardcoded rules for which inlay hint kinds were
prefixes vs suffixes. Explicit location information was added in
https://github.com/llvm/llvm-project/commit/7c19fdd59939249c23384f0900d49aab4a5f0695
so we use this.

We previously generate different "classes" (TextEditorDecorationType) of
decorations for different categories of hints. However we don't actually
render them differently, and VSCode's internal inlay hint API (which we
hope to move to) doesn't have this capability.
Therefore this patch uses a single TextEditorDecorationType for everything.

The TextEditorDecorationTypes were coupled to whether the decorations
were prefixes or suffixes. However this isn't necessary: the
"attachments" are not rendered if empty, so it's enough to set the text
on the decoration itself.

### DIFF
--- a/src/inlay-hints.ts
+++ b/src/inlay-hints.ts
@@ -1,7 +1,6 @@
-// This file implements the client side of the proposed inlay hints
-// extension to LSP. The proposal is based on the one at
-// https://github.com/microsoft/language-server-protocol/issues/956,
-// with some modifications that reflect discussions in that issue.
+// This file implements the client side of clangd's inlay hints
+// extension to LSP: https://clangd.llvm.org/extensions#inlay-hints
+//
 // The feature allows the server to provide the client with inline
 // annotations to display for e.g. parameter names at call sites.
 // The client-side implementation is adapted from rust-analyzer's.
@@ -16,8 +15,8 @@ export function activate(context: ClangdContext) {
   context.client.registerFeature(feature);
 }
 
-// Currently, only one hint kind (parameter hints) are supported,
-// but others (e.g. type hints) may be added in the future.
+/// Protocol ///
+
 enum InlayHintKind {
   Parameter = 'parameter',
   Type = 'type'
@@ -25,6 +24,7 @@ enum InlayHintKind {
 
 interface InlayHint {
   range: vscodelc.Range;
+  position?: vscodelc.Position; // omitted by old servers, see hintSide.
   kind: InlayHintKind|string;
   label: string;
 }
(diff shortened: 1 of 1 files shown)


HighCommander4:
A couple of high-level questions:

 1. Will we want to allow users to turn on/off individual hint kinds (e.g. parameter hints) via client-side flags? (If so, that would be a reason to keep the `kind` around in the longer term.)
 2. Might it make sense to allow users to customize the appearance of hints by hint kind (e.g. have type hints and parameter hints show up in different colors)? (If so, that would be a reason to keep using different TextEditorDecorationTypes for them, though I suppose we could always bring that back later.)

sam-mccall:
Trying to get rid of the different styles was a bad idea.
I'll rework this patch to support position without that.

> Will we want to allow users to turn on/off individual hint kinds (e.g. parameter hints) via client-side flags?

Potentially... We don't have a maintainable way to expose this kind of thing across editors. So I was imagining having a switch to turn the feature on/off (vscode has this) and requiring `config.yaml` for fine-grained configuration.
That said, this is no fun for users. I have a pipe dream that if we had our config schema as data, we could generate a mapping to VSCode settings...

> Might it make sense to allow users to customize the appearance of hints by hint kind 

Yes, definitely. I thought we should drop the styles because I felt certain VSCode inlay hints wasn't going to support them, and that we'd move to that API anyway.
However it looks like this is getting way richer: https://github.com/microsoft/vscode/blob/e8818ffc83a14f24fc081177686fedfbaf055a44/src/vs/editor/common/languages.ts#L1760-L1792

(We'll want to take advantage of this richness on the server eventually...)

sam-mccall:
OK, this should be in better shape.

It now focuses on:
 - respecting `position`
 - supporting arbitrary `kind`s
 - avoiding coupling TextEditorDecorationType to a decision about before/after, so they can be 1:1 with `kind`

src/inlay-hints.ts
```diff
@@ -39,55 +39,67 @@ export const type =
...
+  }
+  return result;
 }
 
-interface HintStyle {
-  decorationType: vscode.TextEditorDecorationType;
-
-  toDecoration(hint: InlayHint,
-               conv: vscodelc.Protocol2CodeConverter): vscode.DecorationOptions;
+function rangeEq(a: vscodelc.Position, b: vscodelc.Position) {
```
HighCommander4, line 58: `positionEq`?

HighCommander4 (approved):
LGTM, thanks

PR #288, opened 2022-01-25
Support cleanup of old version after install

sam-mccall:
Depends on https://github.com/clangd/node-clangd/pull/12


ampandey-AMD (commented):
I am also receiving this error [coc.nvim] Error on notification "runCommand": Command: clangd.install not found.

Does this patch resolves this issue?

PR #291, opened 2022-01-26
Subscribe the registered inlyHints.toggle command.

hokein:
Otherwise it is not released -- we hit a "command 'clangd.inlayHints.toggle' already exists" error
when we run the "restart clangd" command.


sam-mccall:
Whoops, I actually remembered while writing this but then forgot to go back and add it...

PR #296, opened 2022-02-15
Update clang-format

DominicTobias:
I'm hoping this fixes some ignored clang format options such as the ability to have initializer list items on [separate lines](https://stackoverflow.com/questions/56811537/clang-format-how-to-keep-each-element-of-constructors-initializer-list-on-a-se) though tbh the clang-format releases aren't well documented so it's hard to see


sam-mccall:
Thank you!

PR #301, opened 2022-02-25
Use VSCode's native inlay-hints support, now it's available.

sam-mccall:
This isn't ready to go yet:
 - the latest vscode release (1.64) doesn't contain this feature. Even once it does, we maybe don't want to bump our version requirement right away
 - the typings are also not available yet, so here I've included typings for the whole VSCode API (via `vpx vscode-dts main`).

Nevertheless it does work in latest insiders. VSCode seems to request hints for the whole doc, but then continues to issue requests as you scroll around. It makes me a bit sad that scrolling will now eat CPU!

cc @HighCommander4


yashLadha:
I think the PR for native support inlay hints just merged and released in vscode (latest stable version), so we can try to fast-track this PR. 

Thanks for the work @sam-mccall

hokein (approved):
The change looks good, and I test it with the latest vscode 1.66, it works great.

Issue #298, opened 2022-02-22
Error: command 'clangd.inlayHints.toggle' already exists


HighCommander4:
I sometimes get this error when restarting clangd.

```
Error: command 'clangd.inlayHints.toggle' already exists
	at u.registerCommand (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:93:28683)
	at Object.registerCommand (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:107:30423)
	at n9.initialize (/home/nr/.vscode/extensions/llvm-vs-code-extensions.vscode-clangd-0.1.15/out/bundle.js:239:7973)
	at Mb.initializeFeatures (/home/nr/.vscode/extensions/llvm-vs-code-extensions.vscode-clangd-0.1.15/out/bundle.js:36:13023)
	at /home/nr/.vscode/extensions/llvm-vs-code-extensions.vscode-clangd-0.1.15/out/bundle.js:36:5978
```

I think this might happen because this command is registered [at initialize() time](https://github.com/clangd/vscode-clangd/blob/21cc1a5ad494c5f1f9a728c9ee7e1eab5e22a2a9/src/inlay-hints.ts#L133) as opposed to [in the feature constructor](https://github.com/clangd/vscode-clangd/blob/21cc1a5ad494c5f1f9a728c9ee7e1eab5e22a2a9/src/switch-source-header.ts#L7), and `initialize()` gets called again when restarting? (Not sure why it doesn't happen on _every_ restart though...)

HighCommander4:
Ah, it doesn't happen on **explicit** (`clangd.restart` command invocation) restarts because that invokes disposables [here](https://github.com/clangd/vscode-clangd/blob/21cc1a5ad494c5f1f9a728c9ee7e1eab5e22a2a9/src/extension.ts#L22), but it happens on **crash** restarts because... that doesn't?

HighCommander4:
I wasn't sure how to get `clientsOptions.errorHandler` to run disposables (presumably it's possible to wrap the default error handler somehow to do so?), so I opted instead to add a check to `inlay-hints.ts` to ensure we don't try to register the command multiple times.

jwake:
[...]

PR #302, opened 2022-03-01
Avoid registering the 'clangd.inlayHints.toggle' command multiple times

HighCommander4:
initialize() can be called multiple times on a feature (in case the
server crashes and is restarted) without disposables running in
between.

Fixes https://github.com/clangd/vscode-clangd/issues/298


sam-mccall (approved):
Seems ugly that this is required, but it seems to be!

Not going to worry about it too much - the need for this will likely go away as I think VSCode will add a command for this.

vpoughonEZ:
Would you consider releasing 0.1.16 for this?

kleinesfilmroellchen:
Could you, like, ACTUALLY release 0.1.16? It's not very pleasant updating to Clangd 14 and now your entire language server is broken irreperably.

leovincentseles:
Facing with the same problem. Also require a release 0.1.16 too. Thanks

hokein:
I will make a new release.

PR #317, opened 2022-04-01
Fix the broken debugging workflow.

hokein:
The `F5`-debugging workflow has been broken since #287, this patch brings it back.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -21,8 +21,8 @@
         "reveal": "silent",
     },
 
-    // Run custom "compile" script as defined in package.json
-    "args": ["run", "compile", "--loglevel", "silent"],
+    // Run custom "ts-compile" script as defined in package.json
+    "args": ["run", "ts-compile", "--loglevel", "silent"],
 
     // tsc compiler is kept alive and runs in the background.
     "isBackground": true,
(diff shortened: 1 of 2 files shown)


sam-mccall:
Ah, sorry about this.
I don't really understand how this fixes it though - this will build individual source files into *.js, but isn't the package.json going to tell vscode to load the bundled sources anyway?

hokein:
> Ah, sorry about this. I don't really understand how this fixes it though - this will build individual source files into *.js, but isn't the package.json going to tell vscode to load the bundled sources anyway?

not really. The debugging workflow is a combination of `launch.json`, `tasks.json`, and `package.json`:
- `tasks.json` defines the debugging configurations -- we have a `Launch Extension` configuration which is being invoked in `F5` debugging, it has a `"preLaunchTask": "npm"`, the `npm` is defined in `tasks.json` (it runs `npm run compile ...`)
- note that there is a [`problemWatcher`](https://github.com/clangd/vscode-clangd/blob/master/.vscode/tasks.json#L31) settings in `tasks.json`, it watches the tsc-compile output, but after #287, we changed the `npm run compile` from tsc-compile to esbuild-compile, the `problemWatch` setting doesn't match anymore, I think this mismatch causes the debugging failure (where the `Extension Development Host` vscode-window doesn't prompt up).

There is an alternative -- set `problemWatcher` to `$esbuild-watch`, but it requires an extra [extension](https://marketplace.visualstudio.com/items?itemName=connor4312.esbuild-problem-matchers) installed which I want to avoid.

sam-mccall:
Ok that sounds plausible to me and if it works it works!

PR #341, opened 2022-05-30
Fix inlay-hint not shown issue 

### LINKED ISSUE #342: inlay hints gone
It looks like https://github.com/llvm/llvm-project/commit/3137ca80b9ef99359a4adf77512925c2edc461b9 stopped inlay hints from working for me.
`editor.inlayHints.enabled` is on of course. But there are no [`textDocument/inlayHint`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint) messages.
If I comment out https://github.com/clangd/vscode-clangd/blob/dc6518fb0af230c73aa309ef9086e37a49c7e35f/src/inlay-hints.ts#L59 they come back, but based on the `clangd/inlayHints` protocol of course.

Can anyone confirm this behavior?

**System information**

Output of `clangd --version`: `clangd version 15.0.0
Features: linux
Platform: x86_64-unknown-linux-gnu`
Editor/LSP plugin: VSCode 1.67.2 / 0.1.17

hokein:
when using a standard-inlay-hint-supported clangd

- Fixes #342
- Also fixes the `clangd.inlayHints.toggle command not found` issue

### DIFF
--- a/src/inlay-hints.ts
+++ b/src/inlay-hints.ts
@@ -53,7 +53,10 @@ class InlayHintsFeature implements vscodelc.StaticFeature {
     const serverCapabilities: vscodelc.ServerCapabilities&
         {clangdInlayHintsProvider?: boolean, inlayHintProvider?: any} =
         capabilities;
-
+    vscode.commands.executeCommand(
+        'setContext', 'clangd.inlayHints.supported',
+        serverCapabilities.clangdInlayHintsProvider ||
+            serverCapabilities.inlayHintProvider);
     if (!this.commandRegistered) {
       // The command provides a quick way to toggle inlay hints
       // (key-bindable).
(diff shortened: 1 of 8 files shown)


hokein:
should be good to go.

src/inlay-hints.ts
```diff
@@ -53,11 +53,7 @@ class InlayHintsFeature implements vscodelc.StaticFeature {
     const serverCapabilities: vscodelc.ServerCapabilities&
         {clangdInlayHintsProvider?: boolean, inlayHintProvider?: any} =
         capabilities;
-    // If the clangd server supports LSP 3.17 inlay hints, these are handled by
-    // the vscode-languageclient library - don't send custom requests too!
-    if (!serverCapabilities.clangdInlayHintsProvider ||
-        serverCapabilities.inlayHintProvider)
-      return;
+
     if (!this.commandRegistered) {
```
kadircet, line 57: we actually still need to govern entry here. we shouldn't register the command if neither inlay hints extension nor standard protocol are advertised by the server.

kadircet (approved):
thanks, lgtm apart form the command registry condition

src/inlay-hints.ts
hokein: Good point. We should even disable the 'toggle' command in VSCode command palette if users use a clangd without inlay-hints support (v13 or before).

PR #382, opened 2022-09-29
Fix cpptools disabled check to be case-insensitive

appden:
The casing for `C_Cpp.intelliSenseEngine` was recently changed in microsoft/vscode-cpptools#9862 to be lowercase, but this extension will complain every 5 seconds unless the configuration value is strictly the capitalized "Disabled". This uses `toLowerCase()` to still support the legacy capitalized value just like the cpptools extension.


sam-mccall:
Thanks!

PR #386, opened 2022-10-10
[TypeHierarchy] Properly handle standard protocol

kadircet:
Due to a conflict in advertising capability for clangd's type-hierarchy
extension and standard protocol we're going to have both of them available
in newer vscode clients (while clangd's prior to 15 release won't be able to
provide support for the standard protocol).

This patch ensures we're only showing one version of type-hierarchy, preferring
the standard one when clangd supports it, and disabling the standard when clangd
doesn't have support.

### DIFF
--- a/src/type-hierarchy.ts
+++ b/src/type-hierarchy.ts
@@ -101,8 +101,10 @@ class TypeHierarchyTreeItem extends vscode.TreeItem {
 class TypeHierarchyFeature implements vscodelc.StaticFeature {
   private serverSupportsTypeHierarchy = false;
   private state!: vscodelc.State;
+  private context: ClangdContext;
 
   constructor(context: ClangdContext) {
+    this.context = context;
     new TypeHierarchyProvider(context);
     context.subscriptions.push(context.client.onDidChangeState(stateChange => {
       this.state = stateChange.newState;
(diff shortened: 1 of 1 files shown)


src/type-hierarchy.ts
```diff
@@ -116,8 +118,16 @@ class TypeHierarchyFeature implements vscodelc.StaticFeature {
   initialize(capabilities: vscodelc.ServerCapabilities,
              documentSelector: vscodelc.DocumentSelector|undefined) {
     const serverCapabilities: vscodelc.ServerCapabilities&
-        {typeHierarchyProvider?: any} = capabilities;
+        {standardTypeHierarchyProvider?: any, typeHierarchyProvider?: any} =
+        capabilities;
+    // If this is a clangd that supports standard protocol, don't enable the
+    // extension.
+    if (serverCapabilities.standardTypeHierarchyProvider) {
```
sam-mccall, line 125: Nit: the comment doesn't actually mention the unusual circumstance that leads us to do this weird dance.
Without that it's pretty hard to understand the cases. I also think the control flow here (mixing early return with a second condition) is a bit hard to follow. Maybe:

```
// Unfortunately clangd used the same capability name for its pre-standard protocol
// as the standard ended up using. We need to prevent vscode-languageclient from
// trying to query clangd versions that speak the incompatible protocol.
if (typeHierarchyProvider && !standardTypeHierarchyProvider) {
  // clangd speaks pre-standard protocol. Disable library support for the standard protocol.
  ...
} else {
  // Either we're speaking the standard protocol, or there's no implementation at all.
  // In any case, there's nothing to do.
}
```

```

sam-mccall (approved):
Readability suggestions, but feel free to merge in whatever form you're happy with.

PR #387, opened 2022-10-10
Fix an unknown type error

kadircet:


kadircet:
fixes failure in https://github.com/clangd/vscode-clangd/actions/runs/3218520523/jobs/5262792039

hokein (approved):

PR #391, opened 2022-10-12
Remove redundant `TreeDataProvider` registrations

colin-grant-work:
In the type hierarchy tree and AST tree code, both `vscode.window.registerTreeDataProvider` and `vscode.window.createTreeView` are called with the same ID, `clangd.typeHierarchyView` and `clangd.ast`, respectively. These calls are redundant, as `registerTreeDataProvider` just delegates to `createTreeView`:

https://github.com/microsoft/vscode/blob/6f7820c454195fd9d9749022f911a08a004f34bd/src/vs/workbench/api/common/extHostTreeViews.ts#L81-L84

VSCode handles the redundancy gracefully, but when the plugin is run in Theia, errors are logged about both views:

image

image

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -47,7 +47,6 @@ class ASTFeature implements vscodelc.StaticFeature {
             // @ts-ignore
             tree.reveal(null);
         }),
-        vscode.window.registerTreeDataProvider('clangd.ast', adapter),
         // Create the "Show AST" command for the context menu.
         // It's only shown if the feature is dynamicaly available (package.json)
         vscode.commands.registerTextEditorCommand(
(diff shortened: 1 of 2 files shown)


sam-mccall:
Huh, not sure how I confused myself into thinking both were required. Thanks!

HighCommander4:
> Huh, not sure how I confused myself into thinking both were required. Thanks!

Pretty sure the fault lies with me, as I [wrote](https://github.com/clangd/vscode-clangd/pull/20) the type hierarchy view originally. I started with just `registerTreeDataProvider()`, but then I needed a `TreeView` to call `reveal()` on and, not realizing that `registerTreeDataProvider()` created one, I created one directly. The AST view then probably cargo-culted the code from the type hierarchy view :)

PR #392, opened 2022-10-13
Enable clangd on HLSL sources

llvm-beanz:
Clang is gaining support for HLSL. This change enables using clangd for the language server for vscode.

I'm unsure if there is a good way to gate this support on a specific version of clang, but when used with clang-15 it produces an error at the beginning of the file noting that compilation failed. It doesn't do anything too terrible if clang can't handle HLSL, so it might be safe to enable everywhere as-is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
         "onLanguage:cuda-cpp",
         "onLanguage:objective-c",
         "onLanguage:objective-cpp",
+        "onLanguage:hlsl",
         "onCommand:clangd.activate",
         "onCommand:clangd.install",
         "onCommand:clangd.update"
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -18,6 +18,7 @@ export const clangdDocumentSelector = [
   {scheme: 'file', language: 'cuda-cpp'},
   {scheme: 'file', language: 'objective-c'},
   {scheme: 'file', language: 'objective-cpp'},
+  {scheme: 'file', language: 'hlsl'},
 ];
 
 export function isClangdDocument(document: vscode.TextDocument) {


i-ky:
I think you could also advertise HLSL in the extension manifest:
https://github.com/clangd/vscode-clangd/blob/687314dcfaa2ff982bd2aa448ec5db544c5dd77b/package.json#L18-L24

src/clangd-context.ts
```diff
@@ -18,6 +18,7 @@ export const clangdDocumentSelector = [
   {scheme: 'file', language: 'cuda-cpp'},
   {scheme: 'file', language: 'objective-c'},
   {scheme: 'file', language: 'objective-cpp'},
+  {scheme: 'file', language: 'hlsl'},
```
sam-mccall, line 21: this part is "just code", I think we can make it conditional on a config option.

package.json
```diff
@@ -28,6 +28,7 @@
         "onLanguage:cuda-cpp",
         "onLanguage:objective-c",
         "onLanguage:objective-cpp",
+        "onLanguage:hlsl",
```
sam-mccall, line 31: this can't be conditional, but I *think* this just wakes up the plugin when an HLSL file is opened, and would just waste a few resources if the documentSelector doesn't have HLSL.

sam-mccall:
Sorry, I keep losing my comment...

I think we should get this in front of HLSL clang devs ASAP, and in front of "regular" users as soon as there's been an LLVM release with language support in a good state.

At the moment I think all released versions of clangd will produce spurious errors on ~all valid programs. So I think it's too soon for it to be on by default, but we should be able to put it behind a setting.

I'm not too worried about gating on clangd version, we have update/install prompts in the plugin now.

llvm-beanz:
@sam-mccall Are there any other changes you'd like me to make here?

llvm-beanz:
@sam-mccall & @i-ky, any additional changes I need to make here?

i-ky:
LGTM, but I am not a keyholder here.

llvm-beanz:
@HighCommander4 or @hokein could one of you please review this?

llvm-beanz:
@sam-mccall any chance you can take another look at this?

HighCommander4 (changes requested):
> @HighCommander4 or @hokein could one of you please review this?

I only have a passing familiarity with what HLSL is, but I looked through the discussion in this PR so far, and it seems like Sam is on board with adding this (disabled by default) and the updated patch addresses his comments, so I'm happy to approve this (modulo a comment inline).

Just to check my own understanding here: HLSL is being modelled as a "C family language" in clang, such that clang will build for it an AST represented using the same types (`ASTContext`, `Decl`, etc.) as a C/C++ AST, and this is why it's useful for clangd (whose features mostly operate on those AST node types) to be the language server for HLSL files?

> I'm unsure if there is a good way to gate this support on a specific version of clang

The usual way to gate a client feature on server support is to have the server announce support for the feature in its server capabilities (around [here](https://searchfox.org/llvm/rev/c89058291248aecd7d43373b719aff74706026e7/clang-tools-extra/clangd/ClangdLSPServer.cpp#625); the capability name can be a non-standard extension that the server and client agree on). The client can [check](https://github.com/clangd/vscode-clangd/blob/71338a5617193fc79d9907b301a009f495372d00/src/ast.ts#L76-L80) the capabilities in the `initialize()` method of a `StaticFeature` that it [registers](https://github.com/clangd/vscode-clangd/blob/71338a5617193fc79d9907b301a009f495372d00/src/ast.ts#L10).

I'll leave it up to you if you'd like to do this, either in this patch or a follow-up. I'm also fine with not bothering with this.

llvm-beanz:
> Just to check my own understanding here: HLSL is being modelled as a "C family language" in clang, such that clang will build for it an AST represented using the same types (`ASTContext`, `Decl`, etc.) as a C/C++ AST, and this is why it's useful for clangd (whose features mostly operate on those AST node types) to be the language server for HLSL files?

Yes, HLSL is being added to Clang as a standard language. It has a few additional AST nodes to represent its unique syntaxes and semantics, but they are all derived from Clang `Decl`, `Expr` and `Stmt` base types and integrated in the core Clang AST library.

I'll work on the other changes to address your feedback today.

llvm-beanz:
I've posted a PR (https://github.com/llvm/llvm-project/pull/75633) to add a list of languages to clangd's server capabilities.

Issue #390, opened 2022-10-12
Trigger signature help after accepting a function completion.


MK-Alias:
Goodday,

I really enjoy using clangd. Its a must-have tool for every C++ programmer. However, there is one things that really irritates me. Because just as the title implies, it does not give a hover on auto-completing functions/metods. The default is to insert placeholders. Which honestly is weird for an extreme pedant language like C++ to insert illegal syntax (and also spamming the output on functions with a long agrument list). Fortunally it can be turned of with adding this argument.

`--function-arg-placeholders=0`

This will however break VScode's hover -- which is a way better way to use completing arguments all to getter. So now when auto-completing a function/method it will just type out the function **with paranthesis**. Regardless of howmany arguments it has. If you want to use the hover you then must press backspace to remove the parenthesis and then retype to parenthesis. Which is irritating me on a daily basis. 

**Also clangd is not honoring my parenthesis preference with this!**

A nice solution whoud be to just not insert the parenthesis. This will force me (which is also conform my preference) to type the parenthesis myself and with that triggering the hover. But there are probably other things you can do inside the VScode plugin API to make the hover show if not using `function-arg-placeholders`

Thanks for your time for reading this,
and please make people like me a bit happier in this world by given us the hover!

Thank You!

--
using: clangd version 14.0.6
editor: VScode
arguments:
```json
"clangd.arguments": [
    "--function-arg-placeholders=0",
    "--header-insertion=never",
]
```

sam-mccall:
Plan for this would be:
 - intercept completion items in the middleware (we already do this)
 - find items where the cursor is placed after a signature help trigger character (or maybe do this unconditionally?)
 - set completion item command to `editor.action.triggerParameterHints`

wtdcode:
I have a similar problem like this: https://discourse.llvm.org/t/automatic-parameter-suggestions/2419

Basically, given a function:

```C
int var;

void some_function(int a, int b)
```

type `some<tab>` will soon give:

```c
some_function(<pointer here>int a, int b)
```

typing `v` will become:

```c
some_function(v, int b)
```

However, no autocompletion for `var` pops, which is quite annoying...

sam-mccall:
@wtdcode that is a different issue and I don't think one we can do anything about it, please file a bug against vscode.

I have https://github.com/clangd/vscode-clangd/pull/398/files for the original issue.

wtdcode:
[...]

PR #398, opened 2022-10-22
Trigger signature help when accepting code completions, where appropriate

sam-mccall:
As recommended in https://github.com/microsoft/vscode/issues/78806 This is an incomplete fix but better than nothing. Also files https://github.com/microsoft/vscode/issues/164310 in case VSCode can improve the behavior here.

fixes https://github.com/clangd/vscode-clangd/issues/390

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -122,6 +122,20 @@ export class ClangdContext implements vscode.Disposable {
             // notice until the behavior was in several releases, so we need
             // to override it on the client.
             item.commitCharacters = [];
+            // VSCode won't automatically trigger signature help when entering
+            // a placeholder, e.g. if the completion inserted brackets and
+            // placed the cursor inside them.
+            // https://github.com/microsoft/vscode/issues/164310
+            // They say a plugin should trigger this, but LSP has no mechanism.
+            // https://github.com/microsoft/language-server-protocol/issues/274
+            // (This workaround is incomplete, and only helps the first param).
+            if (item.insertText instanceof vscode.SnippetString &&
+                !item.command &&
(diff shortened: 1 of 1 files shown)


hokein (approved):

src/clangd-context.ts
```diff
@@ -122,6 +122,20 @@ export class ClangdContext implements vscode.Disposable {
...
             // to override it on the client.
             item.commitCharacters = [];
+            // VSCode won't automatically trigger signature help when entering
+            // a placeholder, e.g. if the completion inserted brackets and
+            // placed the cursor inside them.
+            // https://github.com/microsoft/vscode/issues/164310
+            // They say a plugin should trigger this, but LSP has no mechanism.
+            // https://github.com/microsoft/language-server-protocol/issues/274
+            // (This workaround is incomplete, and only helps the first param).
+            if (item.insertText instanceof vscode.SnippetString &&
```
HighCommander4, line 132: Is it intentional that the workaround is not reached if the user has set `"clangd.serverCompletionRanking": false` (which causes us to early-return near the top of this function)?

HighCommander4:
This patch has been reviewed and approved, with no further activity on it for over a year. I'm not aware of any outstanding concerns about it. (I remain curious about the [question I asked](https://github.com/clangd/vscode-clangd/pull/398#discussion_r1270200780), but it's not an issue that needs to block the patch.)

I'm going to go ahead and merge this.

src/clangd-context.ts
HighCommander4: (The likely answer here is that there's no reason the workaround couldn't also apply to the `"clangd.serverCompletionRanking": false` case, we'd just have to reorganize the code a bit to handle that case as well. We can do that in a follow-up patch if anyone asks for it.)

PR #423, opened 2022-12-06
Correctly parse clangd.path containing the user's home folder

valord577:
Signed-off-by: valord577 <valord577@gmail.com>

@sam-mccall

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
     },
     "search.exclude": {
         "out": true // set this to false to include "out" folder in search results
-    }
+    },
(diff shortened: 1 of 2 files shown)


valord577:
I set `clangd.path`: "~/.clangd/bin/clangd". But I got error.

sam-mccall:
`~` isn't part of a path here, but it's rather an expansion feature supported by shells like bash, the same as `$FOO`.

vscode-clangd doesn't support it inside settings (there's no shell here), but does support certain vscode-style expansions like `${env:HOME}` which should work for you here.

Issue #318, opened 2022-04-06
Skip warning message for clanngd.path/clangd.arguments when user trust workspace.


simon8233:
I understand this feature is important under open software development.

I use vscode workspace spec to setup remote index server and user local configure path with different codebase/project to speed up internal user usage

Is it possible to skip the warning message when user chooses "Yes, I trust the authors" on VS Code workspace opening?

or

Is it possible to add configure option (to disable it) in clangd configure of user-level VS Code or $HOME/.clangd

I think the first one is better than the second solution due to user already click trust workspace author 

It will help internal usage cases I think.

HighCommander4:
Looked into this a bit. The vscode feature that prompts you to trust a workspace when opening it is called **Workspace Trust** ([user documentation](https://code.visualstudio.com/docs/editor/workspace-trust), [extension author documentation](https://code.visualstudio.com/api/extension-guides/workspace-trust)).

The clangd feature to prompt you when `clangd.path` or `clangd.arguments` is overridden ([added](https://github.com/clangd/vscode-clangd/commit/395bfe1af61bc0184095b55eb686bf4724836f6c) March 2021) predates Workspace Trust (added in vscode 1.57, [released](https://code.visualstudio.com/updates/v1_57) May 2021). It seems to me that Workspace Trust aims to solve a supserset of the problems that the clangd feature solves; in particular, extensions have a [way](https://code.visualstudio.com/api/extension-guides/workspace-trust#configurations-settings) to list configuration options that should only be respected in trusted workspaces.

So, one potential way forward would be to consider clangd's own prompt feature to be obsoleted by Workspace Trust and remove it. (Note, vscode-clangd trunk [now requires](https://github.com/clangd/vscode-clangd/blob/2e4bd9f581854ee1301df5d0df59ed966a8d1898/package.json#L11) vscode >= 1.65, so we can safely assume Workspace Trust in forthcoming vscode-clangd versions.)

There are two ways this could work:

 * Clangd does not opt into Workspace Trust and therefore is not run at all in Restricted Mode. This is already the case today, but it can be overriden by the user in which case they get the full clangd functionality.
 * Clangd opts into Workspace Trust by declaring [limited](https://code.visualstudio.com/api/extension-guides/workspace-trust#static-declarations) support for Restricted Mode, and using `restrictedConfigurations` to specify that `clangd.path` and `clangd.arguments` should not be respected in Restricted Mode.

[...]

PR #451, opened 2023-01-31
Remove custom secure configuration items and rely on workspace trust from vscode

tru:
See discussion in:
- https://github.com/clangd/vscode-clangd/issues/318
- https://github.com/clangd/clangd/discussions/1488

This PR removes the custom logic and sets the configuration items `clangd.path` and `clangd.arguments` to limited in restricted mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,13 @@
         "url": "https://github.com/clangd/vscode-clangd.git"
     },
     "qna": "marketplace",
+    "capabilities": {
+        "untrustedWorkspaces": {
+            "supported": "limited",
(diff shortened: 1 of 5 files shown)


tru:
@sam-mccall any thoughts on this one?

sam-mccall:
Thanks! This basically looks OK to me - as mentioned in the discussion I don't think VSCode's security model here is great but it probably makes sense to align with it anyway.

tru:
Thanks for merging @sam-mccall - when can we expect a new version release?

Issue #406, opened 2022-10-27
Failed to start and restart clangd


fonqL:
Please describe the problem.

![image](https://user-images.githubusercontent.com/75968297/198311354-1cfaa127-3a07-41a9-a2f5-f8be5abb5798.png)

Simply translate: Command ... resulted in an error ...

But 0.1.22 works well.

**Logs**
No output

**System information**
Clangd version (from the log, or `clangd --version`): 
```
clangd version 14.0.3
Features: windows
Platform: x86_64-pc-windows-msvc
```
clangd extension version:
0.1.23

VScode version:
1.69.1

Operating system:
Win10

HighCommander4:
When this happens, can you open Help --> Toggle Developer Tools, and copy what you see in the Console tab? Hopefully there is a stack trace there that gives more information about this error.

fonqL:
[vscode-app-1667030779566.log](https://github.com/clangd/vscode-clangd/files/9893240/vscode-app-1667030779566.log)

Here are logs, but I can't find something related to clang(d)... Hope it will help you.

fonqL:
After debugging the extension I find my previous config `clangd.path: D:\LLVM\bin\clangd` is regarded as an error, and it is solved by appending `.exe`.
A little weird...
This problem is related to @clangd/install 0.1.16. (When I modified @clangd/install dependency version to 0.1.4 it works well)
The location of the exception is here.
![image](https://user-images.githubusercontent.com/75968297/199916379-502d46e9-feda-4f01-9ed5-40c2f7fcfde5.png)

HighCommander4:
Thanks, I can reproduce the original error now.

Steps to reproduce:

  1. Set `clangd.path` to an invalid path
  2. Close and restart vscode
       ==> At this point, on vscode startup, I get a dialog saying: "The clangd binary '`<path>`' was not found. Would you like to download and install clangd 15.0.3"?
  3. Close that dialog (without installing).
  4. Perform "clangd: Restart language server"

Now I get the error:

> Command 'clangd: Restart language server' resulted in an error (Cannot read properties of undefined (reading 'stop'))

The source of the error is the `this.client.stop()` call [here](https://github.com/clangd/vscode-clangd/blob/c0375489682e331ad06c25b60df86e4f102a6407/src/clangd-context.ts#L180). That code is triggered by 'clangd: Restart language server', and it expects `this.client` to be a valid object, but if clangd was not running to begin with, then it will be undefined.

[...]

PR #461, opened 2023-02-27
Avoid undefined object access in ClangdContext.dispose()

HighCommander4:
this.client may not have been set if activate() failed for some reason (e.g. the clangd executable could not be found and the user opted not to install it via auto-install)

Partially fixes https://github.com/clangd/vscode-clangd/issues/406


hokein (approved):
Thanks!

saly740 (commented):
@DominicTobias

PR #483, opened 2023-05-17
Add hotkey to open cppreference documentation

sr-tream:
This PR add hotkey `Shift+f1` to open documentation on cppreference.com, like in CLion.

I added this feature in clangd extension, because VSCode do not provide API to get symbol info from LSP, and standalone doc extensions can't get symbol scope.

Example:
```cpp
std::string foo;
auto bar = foo.da^ta();
```
For this code, standalone extensions opens doc for `std::data`. With this PR clangd open doc for `std::string::data`

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,6 +279,11 @@
                 "category": "clangd",
                 "title": "Toggle inlay hints",
                 "enablement": "clangd.inlayHints.supported"
+            },
+            {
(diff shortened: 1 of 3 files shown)


sr-tream:
Feature implemented in extension, with provided API (#575)

https://github.com/sr-tream/vscode-clangd-docref

PR #486, opened 2023-05-25
Added handling of predefined variable ${userHome}

### LINKED ISSUE #485: clangd extensions should respect vscode settings variables standards
When I edit the path to the clangd library, I want it to set it up as so:

```json
"clangd.path": "$userHome/Library/Application Support/Code/User/globalStorage/llvm-vs-code-extensions.vscode-clangd/install/**version**/clangd_**version**/bin/clangd"
```

This however does not work as the extension does not seem to recognize the `$userHome` variable (see more [here](https://code.visualstudio.com/docs/editor/variables-reference))

For now, using the `$userHome` gives us the path of current working directory and appends the path corresponding to the `clangd.path` setting in **settings.json**. Which means this : 
```json
"CURRENT_WORKING_DIRECTORY/$userHome/Library/Application Support/Code/User/globalStorage/llvm-vs-code-extensions.vscode-clangd/install/**version**/clangd_**version**/bin/clangd"
```

For now I have to give it the absolute path on my system which is `/Users/USER_NAME/Library/...` instead of `$userHome/Library/` which makes the setting less portable on another system as the user might have a different name all together.

**System information**
clangd version 16.0.2 
Features: mac+grpc+xpc
Platform: x86_64-apple-darwin22.4.0
Operating system: macOS 13.3.1

Singha360:
Resolves #485 
Noticed my silly mistake

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,6 @@
+import {homedir} from 'os';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { homedir } from 'os';
 
 // Gets the config value `clangd.<key>`. Applies ${variable} substitutions.
 export function get<T>(key: string): T {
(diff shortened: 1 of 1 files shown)


HighCommander4:
Thanks! Can you clang-format the patch so it passes the CI check please?

HighCommander4:
@hokein added you as a reviewer, just to make sure you're fine with the general approach of supporting new variables on a case by case basis here

hokein (approved):
Thanks, this looks good to me.

PR #487, opened 2023-05-26
Add integration with cmake-tools extension

### LINKED ISSUE #233: How to find compile_commands.json for Debug/Release build matching cmake tools?
We use [cmake tools](https://github.com/microsoft/vscode-cmake-tools) to build from VSCode. It places `compile_commands.json` in specific folders, with either `Debug` or `Release` in the path. It would be great if the two extensions could work together to resolve the path, is that possible?

I'm not sure what either of the two extensions would need to do to make that work. I guess vscode-clangd would need an environment variable that resolves to `Debug` or `Release`? And when switching the build target in cmake tools, vscode-clangd would need to get some notification that the path changed?

sr-tream:
This PR add integration with [cmake-tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) extension. It uses exported API instead of querying the extension instance directly, so this integration must work with any cmake-tools forks, that implement this API.

This PR is cannot use as a complete replacement for `.clangd` configure file - it not pass `--query-driver` to clangd because clangd doesn't support change this via LSP. So this PR just alternative to export compile commands (clangd in VSCode now may work with `--compile_args_from=lsp` flag).

Resolves #233, #498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## Version 0.1.27: March 16, 2023
+## Version 0.1.27: March 16, 2024
 
 * Trigger signature help when accepting code completions, where appropriate [#390](https://github.com/clangd/vscode-clangd/issues/390)
 * Gracefully handle `clangd.restart`` being invoked when the extension hasn't been activated yet [#502](https://github.com/clangd/vscode-clangd/issues/502)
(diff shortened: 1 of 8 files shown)


sr-tream:
In force-push, added missed field in `package-lock.json`

sr-tream:
Feature implemented in extension, with provided API (#575)

https://github.com/sr-tream/vscode-clangd-cmake

PR #509, opened 2023-07-20
Restrict 'Show AST' menu to supported files only

chouzz:
Just found "Show AST" menu will be shown in unrelated places like vscode output channel or even json file after clangd activated.

This patch restricts 'Show AST' menu to supported files only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -303,7 +303,7 @@
                 },
                 {
                     "command": "clangd.ast",
-                    "when": "clangd.ast.supported"
+                    "when": "(resourceLangId == c || resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp) && clangd.ast.supported"
                 }
             ],
             "view/title": [


chouzz:
Hi, I believe it's a useful improvement if you can take a look at this patch. [@sam-mccall](https://github.com/sam-mccall) [@kadircet](https://github.com/kadircet)

HighCommander4 (approved):
Thanks!

It would be nice to reuse the check between `clangd.switchheadersource` and `clangd.ast` (rather than repeat the list of languages) but I'm not sure how to do that. If someone knows, it can be improved in a follow-up.

chouzz:
> Thanks!
> 
> It would be nice to reuse the check between `clangd.switchheadersource` and `clangd.ast` (rather than repeat the list of languages) but I'm not sure how to do that. If someone knows, it can be improved in a follow-up.

I checked about this, seems I can't reuse the checks, at least vscode don't provide a way to do this. But maybe I can make it shorter like cpptools: https://github.com/microsoft/vscode-cpptools/blob/main/Extension/package.json#L3224

HighCommander4:
I think this is fine for now, using a regex would make it a bit shorter but less clear (especially as ours would have to be more complicated as we also support objective-c).

Issue #512, opened 2023-08-05
Provide a command to disable clangd, when it is hogging CPU or RAM


rnikander:
Sometimes when I edit C++ files in my VS Code project, my laptop becomes unusable, VS Code barely responds to mouse and keyboard. I look at the macOS "Activity Monitor" app and see it is `clangd`, hogging the entire CPU and making my fans spin. Sometimes consumes 20GB of RAM, and processes need to be killed, but that's another bug. 

Obviously these are problems with `clangd`, but it would be helpful if there were a VS Code command to temporarily shut it off when editing certain files. This would allow me to get work done. It's not ideal, but it's about a more "graceful failure".

HighCommander4:
Should be straightforward to implement, we already have a `clangd: Restart language server` command which stops and restarts the language server, the new command would do just the stopping part.

PR #524, opened 2023-08-29
Implement clangd.shutdown to manually deactivate the extension (complement to clangd.activate; first half of clangd.restart). Addresses #512

dilijev:
This change is a proof of concept that builds the fix roughly as described in #512 and works to disable and enable clangd language server on command with clangd.shutdown and clangd.activate. clangd.shutdown is implemented with the "dispose" action used for clangd.restart, as proposed in the issue.

This still needs work to add some state checking of whether the language server is already shutdown, unregistration of commands that won't work if the language server is not running, or early-exit from said commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -203,6 +203,11 @@
                 "title": "Manually activate extension"
             },
             {
+                "command": "clangd.shutdown",
+                "category": "clangd",
+                "title": "Manually shutdown extension"
+            },
+            {
                 "command": "clangd.restart",
                 "category": "clangd",
                 "title": "Restart language server"
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -16,7 +16,13 @@ export async function activate(context: vscode.ExtensionContext) {
   // An empty place holder for the activate command, otherwise we'll get an
   // "command is not registered" error.
   context.subscriptions.push(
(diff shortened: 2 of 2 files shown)


dilijev:
Draft notes (TODO remove):

This still needs work to add some state checking of whether the language server is already shutdown, unregistration of commands that won't work if the language server is not running, or early-exit from said commands such that they don't display scary errors.

Certain commands may imply activation of the language server. Depending on the use case, this may or may not make sense versus asking the user to restart the server manually (with command `>clangd: Manually activate extension`) because it was previously manually shutdown by the user. Reactivating the language server via besign actions like going to another C++ file would defeat the benefit, but running clangd extension commands is enough of a signal that it might be reasonable to "do the right thing" by reactivating the language server.

(FWIW, the scenario that prompted me to work on this for example is the desire to temporarily save some system resources while running an intensive build & link task, for example, without having to close out of VS Code, as I can continue to read code and plan my next change while the build is running.)

HighCommander4:
Apologies that this fell off my radar. Another PR has been submitted recently at https://github.com/clangd/vscode-clangd/pull/866; please feel free to share your thoughts there.

PR #531, opened 2023-09-14
Label "close" commands for views; improve enablement

colin-grant-work:
At the moment, the `Close` commands for each contributed view:
 - All have the same title `"Close"`
 - Are all enabled and visible in the command palette even if the corresponding view isn't actually visible

This PR
 - Adds details about the view to be closed to the title so that each is distinct
 - Uses a `${viewId}.visible` context key that is [automatically set up by VSCode](https://github.com/microsoft/vscode/blob/bccfade64adb249f57c8fcf03cba41609f76ce5c/src/vs/workbench/services/views/browser/viewDescriptorService.ts#L892) for each view to determine enablement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -230,7 +230,8 @@
             {
                 "command": "clangd.typeHierarchy.close",
                 "category": "clangd",
-                "title": "Close",
+                "enablement": "clangd.typeHierarchyView.visible",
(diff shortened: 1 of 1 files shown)


HighCommander4 (approved):
Thanks!

PR #544, opened 2023-10-24
Allow setting compilation database directory

wolfpld:
This change adds a configuration entry that allows you to specify a custom location for the `compile_commands.json` file. For example, this might be necessary if you are using the VS Code Meson extension, which by default uses `builddir` as the build directory containing the above file.

Or imagine you have a single workspace where you have two separate build directories with different feature flags enabled. Perhaps you are targeting multiple platforms in this way? You want to be able to quickly switch between these different configurations.

While you can achieve the same result by adding `--compile-commands-dir=...` to the `clangd.arguments` configuration variable, this is hardly discoverable by users. Adding a separate variable with a proper description will better expose the functionality.

Also, while `clangd.compilationDatabaseDirectory` is currently restricted to trusted workspaces only, a proper check on the validity of the specified path would allow this restriction to be removed, which cannot be said for the more general `clangd.arguments` parameter. Unfortunately, the VS Code API does not seem to expose any functionality to do this, and I am not able to do it safely myself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,6 +170,11 @@
                     "type": "number",
                     "default": 0.55,
                     "description": "Opacity of inactive regions (used only if clangd.inactiveRegions.useBackgroundHighlight=false)"
+                },
+                "clangd.compilationDatabaseDirectory": {
+                    "type": "string",
+                    "default": "",
+                    "decription": "Workspace-relative directory to search for a compilation database compile_commands.json file. Leave empty to use the workspace root."
                 }
             }
         },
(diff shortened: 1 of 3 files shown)


tristan957:
For what it is worth, you can pass the directory containing the compile_commands.json to clang through the `clangd.arguments` setting.

wolfpld:
Yes, this is what I want to work on today. The downside is that we will overwrite any custom settings the user may have provided.

tristan957:
@wolfpld I agree that settings the compilation database is a little bit more special than the other arguments.

Issue #501, opened 2023-07-10
Expose AST for use in other extensions


meisZWFLZ:
I'm asking for a non-user-facing command that would provide the same information as the `Show AST` command, but rather than displaying the AST to the user, the command would instead return the AST.

This would be useful for extensions that would enjoy a greater understanding of the code than vscode's symbol interface. For example, I am creating a vscode extension that interprets function calls into a graphical representation, which can be modified to edit the underlying function call (these function calls are robot instructions).

HighCommander4:
This is an interesting idea.

I don't think vscode-clangd currently exposes any API surface meant for consumption by other vscode extensions, but I believe that sort of thing is possible to do.

I think a first step here would be to come up with a proposed API.

meisZWFLZ:
Alright, I believe there is two different methods to do this:

1. Expose via command - Register a command called "clangd.ast.retrieve" that takes a `Range` and `Uri` as parameters and returns an `ASTNode` (I have already implemented [this](https://github.com/meisZWFLZ/vscode-clangd-expose-ast)) 
2. Expose via `Extension.exports` - In the `activate()` function, return an object containing `astRetrieve()` with a signature similar to that of the command (There is no documentation/examples on this that I can find, aside from the Extension class [docs](https://code.visualstudio.com/api/references/vscode-api#Extension&lt;T&gt;))

I don't know if either method is any better than the other, but I'd be happy to implement either and submit a pull request.
 
Also, how could it be made easier to find this api for other extension developers, or is this issue enough?

HighCommander4:
I had in mind the second approach. I think this is more flexible, as it gives us the ability to grow the extension's public API over time as needed, rather than being limited to commands.

> Also, how could it be made easier to find this api for other extension developers, or is this issue enough?

We should document the extension's public API once it has one. Not sure where yet; https://clangd.llvm.org/ is a possibility, but I'm not sure if having client-specific documentation there is appropriate. A wiki page in this repo is another option.

meisZWFLZ:
I added the `ExtensionAPI` type and `ClangdContext.registerAPIFeature()` method, such that any future APIs may be implemented by adding an entry to `ExtensionAPI` and registering it with the ClangdContext.

Two things I noticed while implementing this with my consuming extension:
### Type Library? 
I need to use the types in this repository; is there a way to export and import these types in a way that makes sense (and isn't copying and pasting)

### onDidModifyAST Event 
[...]

PR #575, opened 2023-12-29
Export client from activation as API for use in other extensions

thegecko:
This PR exports the language client so other VS Code extensions can take advantage of the clangd LSP features.

For example, domain specific functionality (e.g. intrinsics) could be implemented in a separate extension and trivially leverage the full-featured AST functionality exposed by clangd.

This has been tested and working with a trivial hover example:

```typescript
import * as vscode from 'vscode';
import { BaseLanguageClient, Range, TextDocumentIdentifier, RequestType } from 'vscode-languageclient';
import { ClangdExtension } from ''

const CLANGD_EXTENSION = 'llvm-vs-code-extensions.vscode-clangd';
const CLANGD_API_VERSION = 1;

interface ClangdApiV1 {
  languageClient: BaseLanguageClient
}

interface ClangdExtension {
  getApi(version: 1): ClangdApiV1;
}

interface ASTParams {
    textDocument: TextDocumentIdentifier;
    range: Range;
}

interface ASTNode {
    role: string;    // e.g. expression
    kind: string;    // e.g. BinaryOperator
    detail?: string; // e.g. ||
    arcana?: string; // e.g. BinaryOperator <0x12345> <col:12, col:1> 'bool' '||'
    children?: Array<ASTNode>;
    range?: Range;
}

// This should work, but an error about kind: auto is thrown :/
// const ASTRequestType = new RequestType<ASTParams, ASTNode|null, void>('textDocument/ast');
const AST_REQUEST_TYPE = 'textDocument/ast';

export class LanguageFeatures {
    public constructor(protected context: vscode.ExtensionContext) {
        context.subscriptions.push(
            vscode.languages.registerHoverProvider(['c', 'cpp'], {
                provideHover: (document, position, token) => this.provideHover(document, position, token)
            })
        );
    }

    protected async provideHover(document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> {
        try {
            const languageClient = await this.getLanguageClient();
            if (!languageClient) {
                return undefined;
            }

            const textDocument = languageClient.code2ProtocolConverter.asTextDocumentIdentifier(document);
            const range = languageClient.code2ProtocolConverter.asRange(new vscode.Range(position, position));
            const params: ASTParams = { textDocument, range };
            const ast: ASTNode | undefined = await languageClient.sendRequest(AST_REQUEST_TYPE, params);

            if (!ast) {
                return undefined;
            }

            return {
                contents: [ast.kind]
            };
        } catch {
            // Ignore any errors
        }
    }

    protected async getLanguageClient(): Promise<BaseLanguageClient | undefined> {
        const extension = vscode.extensions.getExtension(CLANGD_EXTENSION);
        if (!extension) {
            return undefined;
        }

        const activated = await extension.activate() as ClangdExtension;
        const api = activated.getApi(CLANGD_API_VERSION);
        return api.languageClient;
    }
}

```

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -9,9 +9,9 @@ import type { ClangdExtension, ASTParams, ASTNode } from '@clangd/vscode-clangd'
 const CLANGD_EXTENSION = 'llvm-vs-code-extensions.vscode-clangd';
 const CLANGD_API_VERSION = 1;
 
-const ASTType = 'textDocument/ast';
+const ASTRequestMethod = 'textDocument/ast';
 
-const provideHover = (document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> => {
+const provideHover = async (document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> => {
 
     const clangdExtension = vscode.extensions.getExtension<ClangdExtension>(CLANGD_EXTENSION);
 
@@ -22,11 +22,13 @@ const provideHover = (document: vscode.TextDocument, position: vscode.Position, 
         const range = api.languageClient.code2ProtocolConverter.asRange(new vscode.Range(position, position));
         const params: ASTParams = { textDocument, range };
  
-        const ast: ASTNode | undefined = await api.languageClient.sendRequest(ASTType, params);
+        const ast: ASTNode | undefined = await api.languageClient.sendRequest(ASTRequestMethod, params);
 
         return {
             contents: [ast.kind]
         };
     }
 };
+
+vscode.languages.registerHoverProvider(['c', 'cpp'], { provideHover });
 ```
--- a/api/README.md
+++ b/api/README.md
@@ -1,0 +1,30 @@
+# VS Code clangd Extension API
+
+The VS Code clangd extension exposes an API that other extensions can consume:
+
+```typescript
+import * as vscode from 'vscode';
+import type { ClangdExtension, ASTParams, ASTNode, ASTType } from '@clangd/vscode-clangd';
+
+const CLANGD_EXTENSION = 'llvm-vs-code-extensions.vscode-clangd';
+const CLANGD_API_VERSION = 1;
+
+const provideHover = (document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> => {
+
+    const clangdExtension = vscode.extensions.getExtension<ClangdExtension>(CLANGD_EXTENSION);
+
+    if (clangdExtension) {
+        const api = (await clangdExtension.activate()).getApi(CLANGD_API_VERSION);
+ 
+        const textDocument = api.languageClient.code2ProtocolConverter.asTextDocumentIdentifier(document);
+        const range = api.languageClient.code2ProtocolConverter.asRange(new vscode.Range(position, position));
+        const params: ASTParams = { textDocument, range };
+ 
+        const ast: ASTNode | undefined = await api.languageClient.sendRequest(ASTType, params);
+
+        return {
+            contents: [ast.kind]
+        };
+    }
+};
+```
--- a/api/README.md
+++ b/api/README.md
@@ -4,10 +4,12 @@ The VS Code clangd extension exposes an API that other extensions can consume:
 
 ```typescript
 import * as vscode from 'vscode';
-import type { ClangdExtension, ASTParams, ASTNode, ASTType } from '@clangd/vscode-clangd';
+import type { ClangdExtension, ASTParams, ASTNode } from '@clangd/vscode-clangd';
 
 const CLANGD_EXTENSION = 'llvm-vs-code-extensions.vscode-clangd';
 const CLANGD_API_VERSION = 1;
+
+const ASTType = 'textDocument/ast';
 
 const provideHover = (document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> => {
 
--- a/api/README.md
+++ b/api/README.md
@@ -24,6 +24,10 @@ const provideHover = async (document: vscode.TextDocument, position: vscode.Posi
  
         const ast: ASTNode | undefined = await api.languageClient.sendRequest(ASTRequestMethod, params);
 
+        if (!ast) {
+            return undefined;
+        }
+
         return {
             contents: [ast.kind]
         };
--- a/api/package.json
+++ b/api/package.json
@@ -1,0 +1,13 @@
+{
+    "name": "@clangd/vscode-clangd",
+    "version": "0.0.0",
+    "description": "API for the llvm-vs-code-extensions.vscode-clangd VS Code extension",
+    "types": "vscode-clangd.d.ts",
+    "publishConfig": {
+        "registry": "https://npm.pkg.github.com"
+    },
+    "repository": "https://github.com/clangd/vscode-clangd",
+    "dependencies": {
+        "vscode-languageclient": "8.0.2"
+    }
+}
--- a/api/package.json
+++ b/api/package.json
@@ -7,6 +7,8 @@
         "registry": "https://npm.pkg.github.com"
     },
     "repository": "https://github.com/clangd/vscode-clangd",
+    "license": "MIT",
+    "homepage": "https://github.com/clangd/vscode-clangd/blob/master/api/README.md",
     "dependencies": {
         "vscode-languageclient": "8.0.2"
     }
--- a/api/vscode-clangd.d.ts
+++ b/api/vscode-clangd.d.ts
@@ -16,8 +16,6 @@ export interface ASTNode {
   range?: vscodelc.Range;
 }
 
-export const ASTType = 'textDocument/ast';
-
 export interface ClangdApiV1 {
   languageClient: BaseLanguageClient
 }
--- a/api/vscode-clangd.d.ts
+++ b/api/vscode-clangd.d.ts
@@ -1,0 +1,27 @@
+import {BaseLanguageClient} from 'vscode-languageclient';
+import * as vscodelc from 'vscode-languageclient/node';
+
+// The wire format: we send a position, and get back a tree of ASTNode.
+export interface ASTParams {
+  textDocument: vscodelc.TextDocumentIdentifier;
+  range: vscodelc.Range;
+}
+
+export interface ASTNode {
+  role: string;    // e.g. expression
+  kind: string;    // e.g. BinaryOperator
+  detail?: string; // e.g. ||
+  arcana?: string; // e.g. BinaryOperator <0x12345> <col:12, col:1> 'bool' '||'
+  children?: Array<ASTNode>;
+  range?: vscodelc.Range;
+}
+
+export const ASTType = 'textDocument/ast';
+
+export interface ClangdApiV1 {
+  languageClient: BaseLanguageClient
+}
+
+export interface ClangdExtension {
+  getApi(version: 1): ClangdApiV1;
+}
--- a/api/vscode-clangd.d.ts
+++ b/api/vscode-clangd.d.ts
@@ -1,12 +1,27 @@
 import {BaseLanguageClient} from 'vscode-languageclient';
 import * as vscodelc from 'vscode-languageclient/node';
 
-// The wire format: we send a position, and get back a tree of ASTNode.
+export interface ClangdApiV1 {
+  // vscode-clangd's language client which can be used to send requests to the clangd language server
+  // Standard requests: https://microsoft.github.io/language-server-protocol/specifications/specification-current
+  // clangd custom requests: https://clangd.llvm.org/extensions
+  languageClient: BaseLanguageClient
+}
+
+export interface ClangdExtension {
+  getApi(version: 1): ClangdApiV1;
+}
+
+// clangd custom request types (type declarations for other requests may be added later)
+
+// textDocument/ast wire format
+// Send: position
 export interface ASTParams {
   textDocument: vscodelc.TextDocumentIdentifier;
   range: vscodelc.Range;
 }
 
+// Receive: tree of ASTNode
 export interface ASTNode {
   role: string;    // e.g. expression
   kind: string;    // e.g. BinaryOperator
@@ -14,12 +29,4 @@ export interface ASTNode {
   arcana?: string; // e.g. BinaryOperator <0x12345> <col:12, col:1> 'bool' '||'
   children?: Array<ASTNode>;
   range?: vscodelc.Range;
-}
-
-export interface ClangdApiV1 {
-  languageClient: BaseLanguageClient
-}
-
-export interface ClangdExtension {
-  getApi(version: 1): ClangdApiV1;
 }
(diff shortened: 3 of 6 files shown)


HighCommander4:
Added some reviewers for feedback.

This has been proposed before in https://github.com/clangd/vscode-clangd/issues/501 with some work in progress towards it, but the approach discussed there was to expose the AST request specifically, rather than exposing the entire language client.

thegecko:
> This has been proposed before in https://github.com/clangd/vscode-clangd/issues/501 with some work in progress towards it...

Thanks, I hadn't seen that

> ...but the approach discussed there was to expose the AST request specifically, rather than exposing the entire language client.

I see, the benefit of exposing the entire client is that all functionality is available through a known VS Code type (BaseLanguageClient) and there is less need to maintain a hand-crafted interface.

There are 2 potential areas of change for this PR, but I wanted to keep it simple initially.

- Exposing the client directly effectively "blocks" the extension from exposing other APIs in future. This can be changed by instead returning an object containing the client. This approach really requires proper typing

- On types, a common approach is to publish a versioned API package (either on npm or GitHub) exposing the types, with the versioning future proofing API users. I've done this before, so happy to do that here if we see this as the right approach.

If exposing the client (rather than a custom AST API), I would suggest the API includes the RequestTypes as well as the shape of the API (e.g. `ASTRequestType`).

thegecko:
I've updated this PR to add a versioned API which can be extended (the languageClient no longer stomps the root export).

I would appreciate some feedback on getting this merged @HighCommander4 @sam-mccall @kadircet @hokein

HighCommander4:
> I would appreciate some feedback on getting this merged @HighCommander4 @sam-mccall @kadircet @hokein

I'm broadly supportive of making vscode-clangd make its functionality available to other vscode extensions.

I do think this is a significant enough decision for the project that I'm not comfortable making it on my own, and would like one of the other mentioned folks to weigh in as well.

HighCommander4:
I think a link to a real-world example usage of the proposed API (perhaps in a branch of another repo) would help reason about it.

thegecko:
> I think a link to a real-world example usage of the proposed API (perhaps in a branch of another repo) would help reason about it.

Our usage of this is not OSS to link to, but essentially exposing this access means this AST can be leveraged and enhanced functionality can be offered by another vs code extension without having to load clangd twice. Simply put, if a user needs access to a C/C++ AST for adding better IDE experience, this PR offers a simple approach to doing this.

Our use cases are primarily around accessing the AST, but it is cleaner, easier and more extensible to expose the language client in case there are other needs in future.

AST use cases include:

- adding support for target-specific instructions without having to ship a specific clangd flavour
- adding support for documentation links in a custom context e.g. to point at docs for new instructions being exposed
- I assume the request in #501 was a specific need to

HighCommander4:
> > I think a link to a real-world example usage of the proposed API (perhaps in a branch of another repo) would help reason about it.
> 
> Our usage of this is not OSS to link to, but essentially exposing this access means this AST can be leveraged and enhanced functionality can be offered by another vs code extension without having to load clangd twice. Simply put, if a user needs access to a C/C++ AST for adding better IDE experience, this PR offers a simple approach to doing this.

My question was more geared towards understanding what the code consuming the API will look like.

I do see that there is a code example in the PR description -- is that what consuming code would actually look like? For example, it would have to provide its own definition of `interface ASTNode` and `const ASTRequestType`?

> AST use cases include:
>    * adding support for target-specific instructions without having to ship a specific clangd flavour
>    * adding support for documentation links in a custom context e.g. to point at docs for new instructions being exposed

Not sure I'm following these particular examples -- by "instructions", do you mean assembly instructions? (Would that come up in the context of editing inline assembly code in a C/C++ code file then?)

(This question is mostly out of curiosity. I don't doubt that there are interesting things an extension can do with a C/C++ AST.)

thegecko:
> I do see that there is a code example in the PR description -- is that what consuming code would actually look like? For example, it would have to provide its own definition of interface ASTNode and const ASTRequestType?

In this simplistic form, yes. But it is also common practice to extract this API as typescript types and publish them on GitHub releases or npm. I'm happy to do that too and can look into it if this PR gets merged.

> Not sure I'm following these particular examples -- by "instructions", do you mean assembly instructions? (Would that come up in the context of editing inline assembly code in a C/C++ code file then?)

e.g.

1 A customised clangd is created from a downstream LLVM which includes extra instructions/macros
2 Rather than create another vscode extension to expose this LSP, this extension is used (ppointing at the relevant clangd)
3 We want to add docs links to these additional instructions in the hover text (possible when the AST is exposed in this way)

Another interesting use case would be to expose AST data for consumption by AI engines for code helper inference etc.

This PR has been open for >3 Months, is there any further changes required in order to increase interest in merging it @sam-mccall @kadircet @hokein

HighCommander4:
> > I do see that there is a code example in the PR description -- is that what consuming code would actually look like? For example, it would have to provide its own definition of interface ASTNode and const ASTRequestType?
> 
> In this simplistic form, yes. But it is also common practice to extract this API as typescript types and publish them on GitHub releases or npm. I'm happy to do that too and can look into it if this PR gets merged.

Is it possible to have those type definitions live in this repo (I think that doesn't preclude them being a distinct npm package), and have them added as part of this PR?

> 1 A customised clangd is created from a downstream LLVM which includes extra instructions/macros 2 Rather than create another vscode extension to expose this LSP, this extension is used (pointing at the custom clangd) 3 We want to add docs links to these additional instructions in the hover text (which is possible when the AST is exposed in this way)

In a situation like this, where you're working with a downstream LLVM (which includes clangd in-tree), why not make these hover enhancements on the server side in your downstream clangd?

thegecko:
> Is it possible to have those type definitions live in this repo (I think that doesn't preclude them being a distinct npm package), and have them added as part of this PR?

I've extracted these into a package which can be published to npm/GitHub

> In a situation like this, where you're working with a downstream LLVM (which includes clangd in-tree), why not make these hover enhancements on the server side in your downstream clangd?

Because:

- We don't manage the downstream repo nor have access to it
- My team are primarily typescript/extension engineers and management is easier in the extension
- Additions in clangd are somewhat restricted to the language server protocol (TBF docs links is covered but other use cases may not be). This can be extended with custom calls, but without this extension mechanism, you won't be able to access the LSP extensions anyway

thegecko:
Can I have an update on this, please?

HighCommander4:
In the interest of trying to move this forward, I'm going to add myself as a reviewer, and if we don't hear a strong objection from @sam-mccall, @kadircet, or @hokein, I will take the liberty of approving this.

I want to play around with the patch a bit, including writing a (toy) example plugin that consumes the API to get a feel for what using the API looks like. I have a few other patches in my review queue ahead of this, so it may take me a couple of weeks to get around to that. Your patience is appreciated.

thegecko:
> In the interest of trying to move this forward, I'm going to add myself as a reviewer, and if we don't hear a strong objection from @sam-mccall, @kadircet, or @hokein, I will take the liberty of approving this.
I want to play around with the patch a bit, including writing a (toy) example plugin that consumes the API to get a feel for what using the API looks like. I have a few other patches in my review queue ahead of this, so it may take me a couple of weeks to get around to that. Your patience is appreciated.

Thanks!

HighCommander4:
@thegecko I checked out this branch in order to play around with it; however, it's not building for me.

After running `npm install` and then `npm run package`, I get:

```
$ npm run package

> vscode-clangd@0.1.28 package
> vsce package --baseImagesUrl https://raw.githubusercontent.com/clangd/vscode-clangd/master/

Executing prepublish script 'npm run vscode:prepublish'...

> vscode-clangd@0.1.28 vscode:prepublish
> npm run check-ts && npm run esbuild -- --minify --keep-names


> vscode-clangd@0.1.28 check-ts
> tsc -noEmit -p ./


> vscode-clangd@0.1.28 esbuild
> esbuild ./src/extension.ts --bundle --outfile=out/bundle.js --external:vscode --format=cjs --platform=node --minify --keep-names

✘ [ERROR] Could not resolve "../api/vscode-clangd"

    src/ast.ts:7:42:
      7 │ import {ASTParams, ASTNode, ASTType} from '../api/vscode-clangd';
        ╵                                           ~~~~~~~~~~~~~~~~~~~~~~

1 error
 ERROR  npm failed with exit code 1
```

Apologies if I'm missing something obvious, I'm mainly a C++ dev and I get lost easily in Typescript land.

thegecko:
Thanks @HighCommander4

I added a const to the types at the last minute which isn't supported!

That would have failed straight away if the CI had run, but is now fixed.

I can look at adding the extension instantiation (along with the AST Type const) into the api package, but would prefer that to be a follow up PR.

api/README.md
```diff
@@ -0,0 +1,32 @@
...
+const CLANGD_API_VERSION = 1;
+
+const ASTType = 'textDocument/ast';
+
+const provideHover = (document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> => {
+
+    const clangdExtension = vscode.extensions.getExtension<ClangdExtension>(CLANGD_EXTENSION);
+
+    if (clangdExtension) {
+        const api = (await clangdExtension.activate()).getApi(CLANGD_API_VERSION);
```
HighCommander4, line 19: We should probably discourage consumers unconditionally calling `activate()`, since the clangd extension may already be activated and it doesn't expect `activate()` to be called again in that state.

I believe the right pattern here is for consuming code to:
 * use `clangdExtension.exports` to access the `ClangdExtension` object
 * if it's unsure whether the clangd extension is active yet at that point, then guard the access to `.exports` with a `clangdExtension.isActive` check
    * if the check fails, either bail or call `activate()` at that point

This is what I've done in my [example plugin](https://github.com/HighCommander4/vscode-clangd-api-example/blob/5007b4a1b0252f50308cc9b8cdcda4e6d9d8992e/src/extension.ts#L12-L19), and I'd prefer the README does the same. (Sorry to be picky on this point, I worry that users who may not have much familiarity with VSCode extension may look to this README for how to use the API and assume that what it shows is a good way to do things.)

api/README.md
```diff
@@ -0,0 +1,32 @@
...
+        const range = api.languageClient.code2ProtocolConverter.asRange(new vscode.Range(position, position));
+        const params: ASTParams = { textDocument, range };
+ 
+        const ast: ASTNode | undefined = await api.languageClient.sendRequest(ASTType, params);
+
+        return {
+            contents: [ast.kind]
+        };
+    }
+};
```
HighCommander4, line 31: can we show the `registerHoverProvider()` call in the README, just to make it a more complete example?

HighCommander4 (changes requested):
Ok, I adapted the code from the README to turn it into a complete example plugin at https://github.com/HighCommander4/vscode-clangd-api-example, and I've tried it out and it seems to be working well (it successfully adds the AST node kind to the hover contents provided by vscode-clangd itself).

I think this generally looks great. I have a few minor comments below, after which I think this should be good to go!

thegecko:
Thanks for the feedback @HighCommander4 , comments above.

api/README.md
```diff
@@ -0,0 +1,32 @@
...
+```typescript
+import * as vscode from 'vscode';
+import type { ClangdExtension, ASTParams, ASTNode } from '@clangd/vscode-clangd';
+
+const CLANGD_EXTENSION = 'llvm-vs-code-extensions.vscode-clangd';
+const CLANGD_API_VERSION = 1;
+
+const ASTType = 'textDocument/ast';
+
+const provideHover = (document: vscode.TextDocument, position: vscode.Position, _token: vscode.CancellationToken): Promise<vscode.Hover | undefined> => {
```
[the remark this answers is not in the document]
thegecko: Added

api/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "@clangd/vscode-clangd",
```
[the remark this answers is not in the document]
thegecko: Both added

src/ast.ts
```diff
@@ -4,27 +4,17 @@ import * as vscode from 'vscode';
 import * as vscodelc from 'vscode-languageclient/node';
 
 import {ClangdContext} from './clangd-context';
+import type {ASTParams, ASTNode} from '../api/vscode-clangd';
+
+const ASTType = 'textDocument/ast';
```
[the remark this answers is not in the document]
thegecko: Switched to `ASTRequestMethod ` in the code and example

api/vscode-clangd.d.ts
```diff
@@ -0,0 +1,25 @@
+import {BaseLanguageClient} from 'vscode-languageclient';
+import * as vscodelc from 'vscode-languageclient/node';
+
+// The wire format: we send a position, and get back a tree of ASTNode.
```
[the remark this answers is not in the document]
thegecko: Comments added, also reordered the types to make more sense

api/README.md
thegecko: `vscode.Extension.activate()` is designed to be called multiple times and only calls into the extension's `activate()` method once. Subsequent calls return the resolved extension. This is a common pattern and (just to be sure) I tested this scenario by activating it multiple times with logging in the extension `activate()` function. It worked as expected and only activates once.

api/README.md
thegecko: Added

api/README.md
HighCommander4: Thanks for checking on this, I think this is fine then.

HighCommander4 (approved):
Thanks! The updates look good.

Can you get the clang-format CI check to pass please? Then this should be good to merge.

thegecko:
> Can you get the clang-format CI check to pass please? Then this should be good to merge.

Took me a while to realise the error output is a diff and it wasn't happy with line length. Should now be fixed.

HighCommander4:
Merged. Thanks for your work on this!

If anyone writes publically available plugins that make use of this API, please feel free to mention them in a comment for visibility. At some point we could list such plugins (if they are of general interest) in vscode-clangd's Marketplace page.

sr-tream:
I’m very glad that this PR was accepted - now, instead of supporting my own fork, I just moved everything I needed into separate extensions:
 - [CMake integration](https://github.com/sr-tream/vscode-clangd-cmake)
 - [Open documentation on `Shift+F1`](https://github.com/sr-tream/vscode-clangd-docref)
 - [CodeLens support](https://github.com/sr-tream/vscode-clangd-codelens) for [D91930](https://reviews.llvm.org/D91930)

HighCommander4:
> I’m very glad that this PR was accepted - now, instead of supporting my own fork, I just moved everything I needed into separate extensions:
> 
>    * [CMake integration](https://github.com/sr-tream/vscode-clangd-cmake)
> 
>    * [Open documentation on `Shift+F1`](https://github.com/sr-tream/vscode-clangd-docref)
> 
>    * [CodeLens support](https://github.com/sr-tream/vscode-clangd-codelens) for [D91930](https://reviews.llvm.org/D91930)

Very neat!

>    * [CodeLens support](https://github.com/sr-tream/vscode-clangd-codelens) for [D91930](https://reviews.llvm.org/D91930)

This one still requires a forked server, right?

sr-tream:
> This one still requires a forked server, right?

right

HighCommander4:
Sent out https://github.com/clangd/vscode-clangd/pull/653 so the API appears in a vscode-clangd release

Issue #599, opened 2024-03-19
Regression issue on the new 0.1.27 release.


hokein:
Steps:

1. work in a cpp source file, and introduce some code that make clangd crashes.
2. revert the change, and run `clangd: restart language server` command to restart clangd.

Now with 0.1.27, the restart command in 2) doesn't work -- I don't see a new instance of clangd starts. The https://github.com/clangd/vscode-clangd/pull/587 is a related change, @HighCommander4 .

HighCommander4:
Thanks for filing, I'll investigate.

HighCommander4:
Proposed fix: https://github.com/clangd/vscode-clangd/pull/601

Apologies for the regression.

HighCommander4:
I also filed https://github.com/clangd/vscode-clangd/issues/602 as a follow-up for adding some tests for the clangd.restart command to avoid accidentally introducing regressions like this in the future.

hokein:
Confirm that v0.1.28 release works now. Thanks!

PR #601, opened 2024-03-19
Fix check for clangd.restart being a no-op

HighCommander4:
The intention was to only catch the case where the language server is already starting because running the restart command triggered activation of the plugin, but it was also catching the case where the language server wasn't running at all.

Fixes https://github.com/clangd/vscode-clangd/issues/599


hokein (approved):
Thanks for the investigation and quick fix.

Issue #600, opened 2024-03-19
Inactive Region Highlights Persist Through `clangd.restart`


Baricus:
Since version 0.1.25 (which replaced the old behavior of highlighting regions with the default comment color) the `vscode-clangd` extension does not clear either background highlights or lowered opacity of inactive code before reloading the server.  Background highlights or opacity are re-applied to the code (additively in case of the background highlight).  Additionally, if the inactive code changed during the reload due to using a new `compile_commands.json` or edits to the existing one, code can be highlighted as inactive even while it is active.  

As an example, consider the following C++ program:
```cpp
#include <iostream>

#ifdef A
#define MESSAGE "Howdy A!"
#else
#ifdef B
#define MESSAGE "Hello B!"
#else
#error "A or B must be defined"
#endif
#endif

int main() {
    std::cout << MESSAGE << std::endl;
}
```
and the following (handwritten for simplicity) `compile_commands.json`:
```json
[
    {
        "arguments": [
            "/usr/bin/g++",
            "-DA",
            "main.cpp"
        ],
        "directory": "/path/to/folder/containing/both/files/",
        "file": "main.cpp",
        "output": "a.out"
    }
]
```

When loading a folder containing these two files in VSCode, `vscode-clangd` properly marks the inactive regions:
![image](https://github.com/clangd/vscode-clangd/assets/31926270/8b93f7b9-6226-4638-b75a-9be878ad556e)
If we now change `-DA` to `-DB` in `compile_commands.json` and run `clangd.restart` the highlights from `-DA` persist:
![image](https://github.com/clangd/vscode-clangd/assets/31926270/c5b1ece7-509c-41f2-b578-10f3fc4e71b2)
For completeness, this also works on the "error" branch if we drop the define entirely:
![image](https://github.com/clangd/vscode-clangd/assets/31926270/537b9258-ddfd-4cf1-b7b1-2d7eec24815e)

This behaves similarly if the `clangd.inactiveRegions.useBackgroundHighlight` setings is enabled, but the highlights appear to stack, getting darker as you repeatedly reload, even without changing the `compile_commands.json` file.  

My hunch is that this occurs because there is no code to clear the regions in the [dispose](https://github.com/clangd/vscode-clangd/blob/a3925bb55e4432ada100a6a16b2ac9dcf3da1952/src/inactive-regions.ts#L107) method, but I'm very new to VSCode extension development so I am unsure I'm understanding the root cause properly.  

**Logs**
[...]

PR #608, opened 2024-03-23
Clear All Decorations in `dispose`

Baricus:
This PR fixes issue #600, by calling `dispose` for the `decorationType` member of `InactiveRegionsFeature` objects during their own disposal.  Doing so removes the applied decorations, preventing them from persisting after the the extension is stopped or restarted.

### DIFF
--- a/src/inactive-regions.ts
+++ b/src/inactive-regions.ts
@@ -104,5 +104,8 @@ export class InactiveRegionsFeature implements vscodelc.StaticFeature {
   }
 
   getState(): vscodelc.FeatureState { return {kind: 'static'}; }
-  dispose() {}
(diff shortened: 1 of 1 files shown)


HighCommander4 (approved):
Thanks for the fix! I confirmed with a local test that this fixes #600.

It would be nice in principle to add an automated test for this, but looking at https://github.com/clangd/vscode-clangd/blob/master/test/inactive-regions.test.ts that doesn't seem straightforward to do using the existing mocks, and I feel a bit out of depth with typescript testing to suggest a way to do it that involves new mocks. If someone would like to write a test as a follow-up, please feel free :)

PR #613, opened 2024-04-05
Update decoration type for inactive-regions when settings change

roife:
At present, when changing the 'inactive-regions' settings, users need to restart VSCode for the changes to take effect. This PR makes the extension automatically applies these effects upon detecting user modifications to the settings, enhancing user experience.

### DIFF
--- a/src/inactive-regions.ts
+++ b/src/inactive-regions.ts
@@ -50,26 +50,23 @@ export class InactiveRegionsFeature implements vscodelc.StaticFeature {
     const serverCapabilities: vscodelc.ServerCapabilities&
         {inactiveRegionsProvider?: any} = capabilities;
     if (serverCapabilities.inactiveRegionsProvider) {
-      if (config.get<boolean>('inactiveRegions.useBackgroundHighlight')) {
-        this.decorationType = vscode.window.createTextEditorDecorationType({
-          isWholeLine: true,
-          backgroundColor:
-              new vscode.ThemeColor('clangd.inactiveRegions.background'),
-        });
-      } else {
-        this.decorationType = vscode.window.createTextEditorDecorationType({
-          isWholeLine: true,
-          opacity: config.get<number>('inactiveRegions.opacity').toString()
-        });
-      }
+      this.updateDecorationType();
       this.context.subscriptions.push(
           vscode.window.onDidChangeVisibleTextEditors(
               (editors) => editors.forEach(
                   (e) => this.applyHighlights(e.document.fileName))));
       this.context.subscriptions.push(
           vscode.workspace.onDidChangeConfiguration((conf) => {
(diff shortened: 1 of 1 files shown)


src/inactive-regions.ts
```diff
@@ -90,6 +85,24 @@ export class InactiveRegionsFeature implements vscodelc.StaticFeature {
     this.applyHighlights(filePath);
   }
 
+  updateDeclaratedType() {
```
HighCommander4, line 88: DeclaratedType --> DecorationType

src/inactive-regions.ts
```diff
@@ -90,6 +85,24 @@ export class InactiveRegionsFeature implements vscodelc.StaticFeature {
...
+    if (this.decorationType) {
```
HighCommander4, line 89: nit: this can be written slightly more concisely as `this.decorationType?.dispose();`

src/inactive-regions.ts
```diff
@@ -50,26 +50,21 @@ export class InactiveRegionsFeature implements vscodelc.StaticFeature {
...
       this.context.subscriptions.push(
           vscode.window.onDidChangeVisibleTextEditors(
               (editors) => editors.forEach(
                   (e) => this.applyHighlights(e.document.fileName))));
       this.context.subscriptions.push(
           vscode.workspace.onDidChangeConfiguration((conf) => {
-            if (!conf.affectsConfiguration('workbench.colorTheme'))
+            const inactiveSettingsChanged =
+              conf.affectsConfiguration('clangd.inactiveRegions.useBackgroundHighlight')
+              || conf.affectsConfiguration('clangd.inactiveRegions.opacity');
```
HighCommander4, line 62: should we check `"clangd.inactiveRegions.background"` as well?

HighCommander4 (changes requested):
Thanks, this seems like a nice enhancement

src/inactive-regions.ts
roife: I didn't find this option in the settings👀

HighCommander4: Oh, right, that one is under [colors](https://github.com/clangd/vscode-clangd/blob/6839c067bbcb1a001695048f9087678886e27c46/package.json#L183). So it's probably covered by the `workbench.colorTheme` check then.

HighCommander4 (approved):

HighCommander4 (changes requested):
Whoops, I forgot that the CI only runs if I press the 'Approve' button for it.

Could you format the patch so it passes the clang-format check please? You can see the details of the failure at https://github.com/clangd/vscode-clangd/actions/runs/8586071839/job/23528283548?pr=613.

HighCommander4 (approved):

PR #663, opened 2024-08-01
Add dependabot config

tamird:
This will batch all updates into a single PR once a month.

Dependabot was previously removed in
995639f406735aad111e2e863735b2ec6896a536 but this should be much less
noisy than the previous setup.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


tamird:
@HighCommander4 would you mind giving this a look?

HighCommander4:
I added myself as a reviewer, but it will take me a while to get to, as I'm on travels right now, and I'm not really familiar with these Github Actions type things.

HighCommander4:
@tamird could you elaborate on the motivation for this patch?

Is there a problem you're experiencing as a vscode-clangd user (or developer) that this patch solves for you?

HighCommander4:
> @HighCommander4 sorry, neglected to reply. The motivation is that it's easier to stay current incrementally rather than having to do big-bang bumps like #664.

My thoughts here are similar to the ones in the [node-clangd issue](https://github.com/clangd/node-clangd/pull/31#issuecomment-2756750149): I'd be fine with this if you're also volunteering to handle the monthly updates, otherwise I'd lean towards a more conservative configuration such as excluding major version bumps.

tamird:
Yep, I am volunteering.

tamird:
@kadircet are you able to invite me to this repo as well?

tamird:
Thanks @kadircet! Merging.

Issue #560, opened 2023-12-03
Add l10n support to clangd vscode extension 


jihuayu:
I found that clangd-vscode, as a widely used plugin, does not support l10n. Adding localization support can improve the usability of this plugin for developers whose native language is not English. So, I suggest adding l10n support to this plugin.

Starting from VSCode version 1.73.0, VSCode introduced the l10n API. We can use it directly.

If you think adding l10n support is ok, I am willing to work on it.

### Also see
[VSCode doc about l10n](https://code.visualstudio.com/api/references/vscode-api#l10n)
[VSCode l10n GitHub repo](https://github.com/microsoft/vscode-l10n)
[VSCode extension l10n sample](https://github.com/microsoft/vscode-extension-samples/tree/main/l10n-sample)

HighCommander4:
> @HighCommander4
> ### For Question 1
> 
> Yes, we need to upgrade to version 1.73.0. It was [released in October 2022](https://code.visualstudio.com/updates/v1_73), so it has been around for a while now. I believe most people have already upgraded to this version. If we don't upgrade the version, we can handle localization in the code based on the VSCode version. However, for [the properties, commands written in `package.json`](https://github.com/clangd/vscode-clangd/blob/71338a5617193fc79d9907b301a009f495372d00/package.json#L97-L296), we won't be able to localize them. So, I suggest upgrading to version 1.73.0.

Ok, unless we hear an objection in this thread, I think bumping the minimum vscode version is fine.

Older vscode versions will still be able to load older vscode-clangd plugin versions.

> ### For Question 2
> 
> I have two solutions for this issue.
> 
>     * The first one is to add simple l10n support to node-clangd and set the localized language in the VSCode plugin.
> 
>     * The second solution is to use node-clangd message directly as the localization key in VSCode. For example, if node-clangd returns `"Failed to install clangd"` we can use `"Failed to install clangd"` as the key in the localization file, and provide the corresponding localized string like `{"Failed to install clangd": "the l10n string"}`.
> 
> 
> Each of these methods has its advantages and disadvantages. Which one do you think is better?

Looking at how the node-clangd plugin is structured, I see that there is an existing interface named [UI](https://github.com/clangd/node-clangd/blob/b4abe8970a00ba1c61267f7fab394278817e00da/src/index.ts#L28), which defines method signatures that the respective consumers (vscode-clangd and coc-clangd) implement.

[...]

PR #664, opened 2024-08-01
Modernize

tamird:
- Remove obsolete dependencies.
  - abort-controller; `AbortController` is stable since Node 15.4.0.
- Update dependencies.

Depends on https://github.com/clangd/node-clangd/pull/32.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,10 +27,7 @@
         "onLanguage:cpp",
         "onLanguage:cuda-cpp",
         "onLanguage:objective-c",
-        "onLanguage:objective-cpp",
-        "onCommand:clangd.activate",
-        "onCommand:clangd.install",
-        "onCommand:clangd.update"
+        "onLanguage:objective-cpp"
     ],
     "main": "./out/bundle",
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "homepage": "https://clangd.llvm.org/",
     "icon": "icon.png",
     "engines": {
-        "vscode": "^1.65.0"
+        "vscode": "^1.75.0"
     },
     "categories": [
         "Programming Languages",
@@ -34,7 +34,7 @@
     ],
     "main": "./out/bundle",
     "scripts": {
-        "esbuild": "esbuild ./src/extension.ts --bundle --outfile=out/bundle.js --external:vscode --format=cjs --platform=node",
+        "esbuild": "esbuild ./src/extension.ts --bundle --outfile=out/bundle.js --external:@aws-sdk/client-s3 --external:vscode --format=cjs --platform=node",
         "vscode:prepublish": "npm run check-ts && npm run esbuild -- --minify --keep-names",
         "compile": "npm run esbuild -- --sourcemap",
         "check-ts": "tsc -noEmit -p ./",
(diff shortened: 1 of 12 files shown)


tamird:
@HighCommander4 I've updated this now that @clangd/install 0.1.19 is released.

src/install.ts
```diff
@@ -57,6 +56,13 @@ class UI {
     });
     return Promise.resolve(result); // Thenable to real promise.
   }
+  localize(message: string, ...args: Array<string|number|boolean>): string {
```
HighCommander4, line 59: Ah, nice! I forgot about this. Looks like the person who contributed this support to node-clangd only cared about having it in coc-clangd :laughing: 

This also partially answers my question about the vscode engine version: per the original discussion at https://github.com/clangd/vscode-clangd/issues/560, localization requires 1.73.

(You can add "Fixes https://github.com/clangd/vscode-clangd/issues/560" to the commit message.)

HighCommander4 (commented):
Thanks for the patch!

package.json
```diff
@@ -50,25 +50,24 @@
         "git-clang-format": "git-clang-format --extensions=ts"
     },
     "dependencies": {
-        "@clangd/install": "0.1.17",
-        "abort-controller": "^3.0.0",
-        "vscode-languageclient": "8.0.2"
+        "@clangd/install": "0.1.19",
```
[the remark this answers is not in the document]
tamird: Er, let me know when you've released and I'll update.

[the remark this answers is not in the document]
HighCommander4: Posted https://github.com/clangd/node-clangd/pull/37

tamird:
@HighCommander4 have a look at the CI failure. Seems the great minds behind unzipper decided to add a dynamic require in https://github.com/ZJONSSON/node-unzipper/pull/325 for the AWS SDK for some reason, which is inherently incompatible with esbuild.

tamird:
Ah, I was able to mark it external and that seems to work.

package.json
```diff
@@ -50,25 +50,24 @@
...
-        "ovsx": "^0.3.0",
-        "sinon": "^15.2.0",
-        "typescript": "^4.5.5",
-        "vsce": "^2.7.0",
-        "vscode-test": "^1.3.0"
+        "@types/glob": "^8.1.0",
+        "@types/mocha": "^10.0.7",
+        "@types/node": "^22.0.2",
+        "@types/sinon": "^17.0.3",
+        "@types/vscode": "1.97.0",
```
HighCommander4, line 61: This seems wrong. I think this needs to match `engines.vscode`, because when I try to run `npm run package` locally, I get:

>  ERROR  @types/vscode 1.97.0 greater than engines.vscode ^1.75.0. Either upgrade engines.vscode or use an older @types/vscode version

tamird: Matched the engine version.

HighCommander4 (approved):
Thanks, LGTM! CI is green and a local build is successful as well -- I'll go ahead and merge.

package.json
```diff
@@ -27,14 +27,11 @@
...
-        "onLanguage:objective-cpp",
-        "onCommand:clangd.activate",
-        "onCommand:clangd.install",
-        "onCommand:clangd.update"
+        "onLanguage:objective-cpp"
     ],
     "main": "./out/bundle",
     "scripts": {
-        "esbuild": "esbuild ./src/extension.ts --bundle --outfile=out/bundle.js --external:vscode --format=cjs --platform=node",
+        "esbuild": "esbuild ./src/extension.ts --bundle --outfile=out/bundle.js --external:@aws-sdk/client-s3 --external:vscode --format=cjs --platform=node",
```
kadircet, line 34: why do we need aws-sdk dependency here?

tamird: I believe this is saying that we *don't* bundle this package. I don't remember why I had to add this flag. Try removing it and see?

kadircet: sent out https://github.com/clangd/vscode-clangd/pull/815

PR #669, opened 2024-08-06
Remove unused function call

tamird:


HighCommander4:
For context: prior to https://github.com/clangd/vscode-clangd/commit/db14900a7a3abb1dc6440e3e8369d5aba3fca0a8, this function call had the possible side effect of showing the user a prompt, and we were calling it for this side effect.

Since that commit that side effect no longer exists, and the function call seems to be unnecessary.

Issue #681, opened 2024-09-10
File watcher is not activated: clangd does not refresh open files on compile_commands.json changes


WB2210:
Not sure if to tag as bug or enhancement - 

Open files do not refresh in vscode when compile_commands.json changes, unless the each file is manually closed and re-opened or remembering to run restart clangd server command (Which 99% of the time I forget to do after updating compile_commands.json), this causes confusion when switching targets that have different ifdefs and defines.

I see that clangd extension supports file watch, I tried to debug the extension and discovered that the file watch is never called, when I manually edited the code to force call it, when compile_commands.json changes language server restarts and the files in vscode are refreshed properly, without having to manually perform refresh clangd or close and open each file every time a target is switched.

So what's the solution? Can we have the file watch enabled back? (It fixes the issue) Are all opened files supposed to refresh automatically on vscode without restarting the server?
Thanks

**System information**
Clangd version (from the log, or `clangd --version`): 18.1.3 (tried on version 16 too, same issue)
clangd extension version: 1.29
Operating system: mac

HighCommander4:
I think this is a duplicate of #300, with [this comment](https://github.com/clangd/vscode-clangd/issues/300#issuecomment-1049149308) describing the current state of things.

HighCommander4:
Could you specify what config / code change you're referring to?

(And could you please do it in a comment on #300, to keep the discussion in one place? Thanks!)

PR #688, opened 2024-09-21
[#681] re-enable file watcher

WB2210:
https://github.com/clangd/vscode-clangd/issues/681

Re-enable the functionality of file watcher since clangd's builtin restart functionality does not refresh opened files in vscode, leading to potentially misleading ifdef view

### DIFF
--- a/src/config-file-watcher.ts
+++ b/src/config-file-watcher.ts
@@ -21,9 +21,7 @@ class ConfigFileWatcherFeature implements vscodelc.StaticFeature {
 
   initialize(capabilities: vscodelc.ServerCapabilities,
              _documentSelector: vscodelc.DocumentSelector|undefined) {
-    if ((capabilities as ClangdClientCapabilities)
-            .compilationDatabase?.automaticReload)
-      return;
+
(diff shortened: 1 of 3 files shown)


src/config-file-watcher.ts
```diff
@@ -21,9 +21,7 @@ class ConfigFileWatcherFeature implements vscodelc.StaticFeature {
 
   initialize(capabilities: vscodelc.ServerCapabilities,
              _documentSelector: vscodelc.DocumentSelector|undefined) {
-    if ((capabilities as ClangdClientCapabilities)
-            .compilationDatabase?.automaticReload)
```
HighCommander4, line 25: I'd like us to keep respecting the server's `automaticReload` capability by default.

I was thinking of adding a new config option, like `clangd.onConfigChanged.forceEnable: true`, and only skipping this `automaticReload` check if that option is enabled.

HighCommander4 (changes requested):

HighCommander4 (changes requested):

HighCommander4:
Superseded by #808.

PR #690, opened 2024-09-23
Restrict keybindings to supported files

csoren:
The clangd extension does not restrict the default keybindings for the `clangd.switchheadersource` and `clangd.typeHierarchy` commands to supported language ids. This has the effect of blocking other extensions from using these shortcuts for actions. Using the same shortcut for providing similar functionality for familiarity can be desirable for other extensions.

This PR adds a check for the supported language ids to the `when` clauses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -291,12 +291,12 @@
                 "command": "clangd.switchheadersource",
                 "key": "Alt+o",
                 "mac": "Alt+cmd+o",
-                "when": "editorTextFocus"
+                "when": "(resourceLangId == c || resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp) && editorTextFocus"
             },
             {
                 "command": "clangd.typeHierarchy",
                 "key": "Shift+Alt+t",
-                "when": "editorTextFocus"
+                "when": "(resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp) && editorTextFocus"
             }
         ],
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -291,12 +291,12 @@
...
                 "key": "Alt+o",
                 "mac": "Alt+cmd+o",
-                "when": "editorTextFocus"
+                "when": "(resourceLangId == c || resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp) && editorTextFocus"
             },
             {
                 "command": "clangd.typeHierarchy",
                 "key": "Shift+Alt+t",
-                "when": "editorTextFocus"
+                "when": "(resourceLangId == c || resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp) && editorTextFocus"
```
HighCommander4, line 299: nit: `resourceLangId == c` can be omitted because there is no inheritance and thus no type hierarchies in C

(clangd doesn't currently implement type hierarcy for Objective-C either, but the concept of inheritance does exist in Objective-C and a future version of the server might support type hierarchy for it, so `resourceLangId == objective-c` can be kept)

HighCommander4 (commented):
Thanks for the patch!

package.json
csoren: Good point, I have updated the branch.

HighCommander4 (approved):
Thanks, LGTM

HighCommander4:
Released in 0.1.30.

PR #707, opened 2024-10-21
Adding option to follow symlinks when resolving symbols

cgrinker:
When using clangd in vscode with Bazel — trying to do symbol resolution (Go To Definition, Go To Declaration) will result in a file being opened in the Bazel build intermediate folder. This PR adds a settings flag and middlewares on the symbol resolution functions to follow symlinks back to their actual source file.

This issue has been discussed on the clangd project itself a few years ago. [1](https://github.com/clangd/clangd/issues/197) [2](https://github.com/clangd/clangd/issues/544). From my reading whether or not to follow the symlinks was too deep in the compiler infrastructure itself to change on the LSP server, and that whether or not to follow the symlinks might be situationally dependent. 

I attempted to solve the issue using the `--path-mappings ` flag but it doesn't seem suitable for this situation. 

[Hedron's](https://github.com/hedronvision/bazel-compile-commands-extractor generator `compile_commands.json` also seemed like an opportunity to generate different compile commands but [3](https://github.com/hedronvision/bazel-compile-commands-extractor/issues/195) it doesn't seem like one could get *working* compile commands that aren't pointing into bazel's sandbox directory

Having it here behind a feature flag seems like a reasonable middle ground. Folks that need symlink resolution would be able to check the box, and uncheck it if they were getting funky behavior.

I've created a project [here](https://github.com/cgrinker/vscode-clang-symlink-example) to demonstrate the behavior

## Symbol Resolution
image
## Without Symlink Following
image
## With Symlink Following
image

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs';
 import * as vscode from 'vscode';
 import * as vscodelc from 'vscode-languageclient/node';
 
@@ -104,6 +105,11 @@ export class ClangdContext implements vscode.Disposable {
       // We also mark the list as incomplete to force retrieving new rankings.
       // See https://github.com/microsoft/language-server-protocol/issues/898
       middleware: {
+        provideDeclaration: this.remapDefinitionSymlink.bind(this),
+        provideDefinition: this.remapDefinitionSymlink.bind(this),
+        provideTypeDefinition: this.remapDefinitionSymlink.bind(this),
+        provideImplementation: this.remapDefinitionSymlink.bind(this),
+
         provideCompletionItem: async (document, position, context, token,
                                       next) => {
           if (!config.get<boolean>('enableCodeCompletion'))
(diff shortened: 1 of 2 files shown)


src/clangd-context.ts
```diff
@@ -206,4 +212,65 @@ export class ClangdContext implements vscode.Disposable {
...
+            return new vscode.Location(vscode.Uri.file(p), e.range)
+          }
+
+          return e;
+        }
+      })) as vscode.Declaration
+
+      return res;
+    } else {
+      if (definition.uri.scheme != 'file')
```
cgrinker, line 246: It would have been preferable to use the vscode filesystem api here, but while it has functionality for *checking* if a file is a symlink, it doesn't actually have a function to follow it. Gating behind a file uri scheme should avoid situations where you've got some remote filesystem open and we can't actually follow the symlinks

wudisheng:
It seems that the modification introduced in PR is not working as expected in my environment (listed below), when the follow symlink option is enabled, it resolves the symlink correctly (examined by infomation message) but vscode window still lands inside `_virtual_includes` directory.

Version: 1.95.1 (Universal)
Commit: 65edc4939843c90c34d61f4ce11704f09d3e5cb6
Date: 2024-10-31T05:14:54.222Z
Electron: 32.2.1
ElectronBuildId: 10427718
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.1.0

PR #708, opened 2024-10-25
fix #683 Restore VsCode 1.91 behavior of allowing scripts as executable

### LINKED ISSUE #683: No longer possible to use cmd as clangd executable
The VS Code upgrade to 1.92 contains a new node version (20)
This version seems to have as a side effect that scripts can no longer be used as 'executable' on windows
(see similar issue for cmake: https://github.com/microsoft/vscode-cmake-tools/issues/4037#issuecomment-2331286590)

When using VS Code with 1.91, I can use a clangd.cmd that figures out the clangd version to use (the same as we use to compile). However from VS Code 1.92 this is no longer possible. 

I'm no expert in typescript, though I believe this behavior can be re-enabled in ClangdContext::active by passing "shell: true" as part of the options of vscodelc.Executable.

**Logs**

````
[Error - 11:31:01 AM] Clang Language Server client: couldn't create connection to server.
Error: spawn EINVAL
	at ChildProcess.spawn (node:internal/child_process:421:11)
	at Object.spawn (node:child_process:800:9)
	at c:\Users\JVA\.vscode\extensions\llvm-vs-code-extensions.vscode-clangd-0.1.29\out\bundle.js:39:56906
	at async oc.createConnection (c:\Users\JVA\.vscode\extensions\llvm-vs-code-extensions.vscode-clangd-0.1.29\out\bundle.js:39:12067)
	at async oc.start (c:\Users\JVA\.vscode\extensions\llvm-vs-code-extensions.vscode-clangd-0.1.29\out\bundle.js:39:2940)
````

**System information**

Operating system: Windows

JVApen:
In VsCode 1.72 Node was updated to version 20, before this upgrade, it was possible to execute cmd/bat scripts as executable.

After this update this was suddenly broken.

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -69,7 +69,7 @@ export class ClangdContext implements vscode.Disposable {
     const clangd: vscodelc.Executable = {
       command: clangdPath,
       args: await config.get<string[]>('arguments'),
-      options: {cwd: vscode.workspace.rootPath || process.cwd()}
+      options: {cwd: vscode.workspace.rootPath || process.cwd(), shell: true}
     };
     const traceFile = config.get<string>('trace');
     if (!!traceFile) {


JVApen:
fixes #683

JVApen:
closes #683

HighCommander4:
Looks reasonable to me, using `shell: true` was recommended by a vscode-languageclient maintainer in https://github.com/microsoft/vscode-languageserver-node/issues/1556#issuecomment-2348149230.

I assume the PR title should read "vscode **1.91** behavior" (rather than 1.71)?

Do you know if using `shell: true` also means that on Unix-like systems, `~` in `"clangd.path"` or `"clangd.arguments"` will now be expanded to the user's home directory?

JVApen:
@HighCommander4 I don't have a Linux system at hand to test that. Though my understanding of the option is that it enables scripts. Instead of just regular exes. Inside the script, you have access to the env variables.

As such, by using a clangd.sh with `~/clangd $@` should make it possible to use enable it.

JVApen:
@HighCommander4 What should still happen before this can land?

HighCommander4:
> Do you know if using `shell: true` also means that on Unix-like systems, `~` in `"clangd.path"` or `"clangd.arguments"` will now be expanded to the user's home directory?

I did check on Linux and expansion of `~` does **not** happen in either `"clangd.path"` or `"clangd.arguments"`.

It wasn't obvious to me whether that would be the case, because I went digging in the vscode-languageclient source code to see what it does with the `shell: true` parameter, and found that it forwards it to the nodejs API [`child_process.spawn`](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options), for which the documentation of the parameter says:

> runs `command` inside of a shell

That's a bit vague -- it _could_ well include performing shell expansion of the command / its arguments... but I guess not.

Anyways, I think this patch is fine regardless. I mostly wanted to document these findings for the benefit of anyone in the future who might want to get shell expansion to work or wonders why it doesn't already.

PR #715, opened 2024-11-04
FIX: extension activate failed if there is a space in clangd path

quanzhuo:
The last change `0c2d40d7d6b8092309bd2c880a32e59c71472a80` add `shell: true` to clangd options, which cause the extension activate failed if there is a space in clangd's path.

Especially on windows, clangd installed under 'C:\Program Files\'

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -67,7 +67,9 @@ export class ClangdContext implements vscode.Disposable {
       return;
 
     const clangd: vscodelc.Executable = {
-      command: clangdPath,
+      // Quote the path. With `shell: true`, this is needed
+      // in case the path contains spaces.
+      command: `"${clangdPath}"`,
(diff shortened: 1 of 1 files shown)


JVApen:
Thanks for finding and fixing!

src/clangd-context.ts
```diff
@@ -67,7 +67,7 @@ export class ClangdContext implements vscode.Disposable {
       return;
 
     const clangd: vscodelc.Executable = {
-      command: clangdPath,
+      command: `"${clangdPath}"`,
```
HighCommander4, line 70: Could you add a comment please, something like:

```
// Quote the path. With `shell: true`, this is needed
// in case the path contains spaces.
```

HighCommander4 (changes requested):
Thanks for the patch!

HighCommander4 (approved):

PR #728, opened 2024-11-17
Support API usage when clangd is disabled or failed to initialize properly

HighCommander4:
Fixes https://github.com/clangd/vscode-clangd/issues/721

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -56,25 +56,25 @@ class EnableEditsNearCursorFeature implements vscodelc.StaticFeature {
   dispose() {}
 }
 
-export async function createContext(globalStoragePath: string,
-                                    outputChannel: vscode.OutputChannel):
-    Promise<ClangdContext|null> {
-  const subscriptions: vscode.Disposable[] = [];
-  const clangdPath = await install.activate(subscriptions, globalStoragePath);
-  if (!clangdPath)
-    return null;
-
-  const clangdArguments = await config.get<string[]>('arguments');
-
-  return new ClangdContext(clangdPath, clangdArguments, outputChannel);
-}
-
 export class ClangdContext implements vscode.Disposable {
   subscriptions: vscode.Disposable[] = [];
   client: ClangdLanguageClient;
 
-  constructor(clangdPath: string, clangdArguments: string[],
-              outputChannel: vscode.OutputChannel) {
+  static async create(globalStoragePath: string,
+                      outputChannel: vscode.OutputChannel):
+      Promise<ClangdContext|null> {
+    const subscriptions: vscode.Disposable[] = [];
+    const clangdPath = await install.activate(subscriptions, globalStoragePath);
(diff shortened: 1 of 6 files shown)


HighCommander4:
@thegecko I can't actually add you as a reviewer on this patch, but would appreciate if you could have a look anyways, and see if it looks like a suitable alternative to #727

HighCommander4:
> @thegecko I can't actually add you as a reviewer on this patch

(I guess that's because you're not a project member; and I don't have admin access to the repo to make you a project member. @hokein, perhaps you can help with that?)

src/clangd-context.ts
```diff
@@ -56,19 +56,28 @@ class EnableEditsNearCursorFeature implements vscodelc.StaticFeature {
   dispose() {}
 }
 
+export async function createContext(globalStoragePath: string,
```
thegecko, line 59: minor: You could make this a static async function on the `ClangdContext` class and make the constructor on `ClangdContext` private to force usage through this static factory method.

thegecko (approved):
👍  This looks a lot neater than #727, I've made one minor suggestion.

There are still a couple of non-null assertions which I addressed in #727. These could be picked up in a subsequent PR.

- https://github.com/clangd/vscode-clangd/blob/c7c91e644eec4cd46b62770336fac2a268426675/src/install.ts#L123
- https://github.com/clangd/vscode-clangd/blob/c7c91e644eec4cd46b62770336fac2a268426675/src/clangd-context.ts#L123

src/extension.ts
```diff
@@ -31,20 +30,27 @@ export async function activate(context: vscode.ExtensionContext):
...
+          apiInstance.client = clangdContext?.client;
         }
       }));
 
   let shouldCheck = false;
 
   if (vscode.workspace.getConfiguration('clangd').get<boolean>('enable')) {
-    await clangdContext.activate(context.globalStoragePath, outputChannel);
+    clangdContext =
+        await createContext(context.globalStoragePath, outputChannel);
```
thegecko, line 51: With a static factory method, this would then be:
```suggestion
        await ClangdContext.createContext(context.globalStoragePath, outputChannel);
```

src/clangd-context.ts
HighCommander4: Done; thanks for the suggestion.

HighCommander4:
Thanks!

> There are still a couple of non-null assertions which I addressed in #727. These could be picked up in a subsequent PR.
> 
>     * https://github.com/clangd/vscode-clangd/blob/c7c91e644eec4cd46b62770336fac2a268426675/src/install.ts#L123
> 
>     * https://github.com/clangd/vscode-clangd/blob/c7c91e644eec4cd46b62770336fac2a268426675/src/clangd-context.ts#L123

I went ahead and fixed these.

thegecko:
Thanks for adding those changes, looks great.

thegecko (approved):

PR #730, opened 2024-11-18
Restore VsCode 1.91 behavior of allowing scripts as executable

JVApen:
In VsCode 1.72 Node was updated to version 20, before this upgrade, it was possible to execute cmd/bat scripts as executable.After this update this was suddenly broken.

Fixes clangd#683

In this reapply, we quote both the command line and the arguments, such that both can contain spaces.
We also introduce the option useScriptAsExecutable, such that this is only enabled when the user wants it.

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -68,27 +68,26 @@ export class ClangdContext implements vscode.Disposable {
     if (!clangdPath)
       return null;
 
-    const useScriptAsExecutable =
-        await config.get<boolean>('useScriptAsExecutable');
-    const clangdArguments = await config.get<string[]>('arguments');
-
-    return new ClangdContext(clangdPath, clangdArguments, outputChannel,
-                             useScriptAsExecutable);
+    return new ClangdContext(clangdPath, outputChannel);
   }
 
-  private constructor(clangdPath: string, clangdArguments: string[],
-                      outputChannel: vscode.OutputChannel,
-                      useScriptAsExecutable: boolean) {
+  private constructor(clangdPath: string, outputChannel: vscode.OutputChannel) {
+    const useScriptAsExecutable = config.get<boolean>('useScriptAsExecutable');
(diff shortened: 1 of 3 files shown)


JVApen:
@HighCommander4 The issue with the space was clear and should be resolved. However, I had the impression that there were other reasons why it failed which we don't know about.
As such, I've introduced an option, such that this minimizes the risk of breaking peoples experience once again.

src/clangd-context.ts
```diff
@@ -68,13 +68,23 @@ export class ClangdContext implements vscode.Disposable {
     if (!clangdPath)
       return null;
 
+    const useScriptAsExecutable =
+        await config.get<boolean>('useScriptAsExecutable');
```
HighCommander4, line 72: I simplified this a bit in https://github.com/clangd/vscode-clangd/pull/731 (`config.get` isn't actually async and doesn't need to be awaited), would you mind rebasing on top of that and querying the config value directly in the constructor?

src/clangd-context.ts
```diff
@@ -68,13 +68,23 @@ export class ClangdContext implements vscode.Disposable {
...
 
   private constructor(clangdPath: string, clangdArguments: string[],
-                      outputChannel: vscode.OutputChannel) {
+                      outputChannel: vscode.OutputChannel,
+                      useScriptAsExecutable: boolean) {
+    if (useScriptAsExecutable) {
+      let quote = (str: string) => { return `"${str}"`; };
+      clangdPath = quote(clangdPath)
+      for (var i = 0; i < clangdArguments.length; i++)
+      clangdArguments[i] = quote(clangdArguments[i]);
```
HighCommander4, line 86: nit: indent statement inside loop

JVApen: Completely agree, will look at it

HighCommander4 (approved):
LGTM

PR #731, opened 2024-11-19
Follow-up cleanups after #728

HighCommander4:
- The disposables passed in to install.activate() were not saved for cleanup
 - config.get() is not async

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -57,7 +57,7 @@ class EnableEditsNearCursorFeature implements vscodelc.StaticFeature {
 }
 
 export class ClangdContext implements vscode.Disposable {
-  subscriptions: vscode.Disposable[] = [];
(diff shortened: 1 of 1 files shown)


HighCommander4:
@thegecko or @JVApen perhaps you could have a brief look at this? (once again, I can't actually add you to "Reviewers")

JVApen:
@HighCommander4 Can you keep the arguments as constructor parameter, see also my PR. Beside that, it looks good to me. I am however no expert in Javascript/Typescript

HighCommander4:
> @HighCommander4 Can you keep the arguments as constructor parameter, see also my PR.

Not sure why being a constructor parameter helps; you can always make it a local variable in the constructor.

HighCommander4:
Rebased on top of #730, which took care of the second part (not `await`-ing `config.get`).

Issue #517, opened 2023-08-16
Set up snapshot releases


HighCommander4:
It would be nice to have a way for users to try the latest changes to vscode-clangd without having to build it from source (and without having to publish a new version to the extension marketplace after every commit).

Clangd has weekly snapshot builds that serve this purpose; perhaps vscode-clangd could have something similar?

The frequency of commits to vscode-clangd is pretty low (there have been 6 commits in 2023 so far, that's less than one per month on average), so it could make sense to have every commit trigger a snapshot build (but I'm open to alternatives, e.g. a daily or weekly snapshot that's only created if there have been changes since the last snapshot).

GitMensch:
A direct build seems also to be much more reasonable for this extension - as it is much smaller an faster to do than the full clangd.
I'd therefore "vote" for integrating that in the GH workflow.

simon8233:
is it any exists vscode-clangd release docs for developer refers?
I hope i can found some spase time to do some contribute on it 
vscode-clangd give much help on my jobs :) 
very thanks community help:)

HighCommander4:
> is it any exists vscode-clangd release docs for developer refers? I hope i can found some spase time to do some contribute on it vscode-clangd give much help on my jobs :) very thanks community help:)

Thanks for your interest :) You can find instructions for building vscode-clangd here: https://github.com/clangd/vscode-clangd/blob/master/DEVELOPING.md

HighCommander4:
Update for anyone following this issue: I've added a Github workflow for creating an unstable release (which does not get published to the VSCode Marketplace or OpenVSX) in https://github.com/clangd/vscode-clangd/pull/748.

Such releases currently need to be manually triggered. As a next step, we likely want to trigger them automatically.

I'm open to opinions on how frequently they should be automatically triggered. Weekly, like clangd server snapshot releases? (There will be weeks when there are no commits, I guess we should skip those?) Or maybe on every commit, since the volume of commits in this project is not that high?

PR #748, opened 2024-12-06
Add a Github workflow for creating an unstable release

HighCommander4:
The workflow is manually triggered, and creates a Github release containing a build from the master branch.

### DIFF
--- a/.github/workflows/unstable-release.yml
+++ b/.github/workflows/unstable-release.yml
@@ -1,0 +1,38 @@
+name: Publish unstable release
+on:
+  workflow_dispatch: 
+    inputs:
+      tag:
+        description: 'Tag name'
+        required: true
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - name: start X server
+      run: |
+        Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
+        echo "DISPLAY=:99" >> $GITHUB_ENV
+    - uses: actions/checkout@v2
+    - name: read version
+      id: version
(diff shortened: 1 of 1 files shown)


HighCommander4:
This is a first step towards https://github.com/clangd/vscode-clangd/issues/517. So far the action is manually triggered, but we could consider expanding this in a future change to a regularly scheduled snapshot release, similar to what the clangd server has.

HighCommander4:
Example of the workflow running successfully in my fork: https://github.com/HighCommander4/vscode-clangd/releases/tag/snapshot_20241205

.github/workflows/unstable-release.yml
```diff
@@ -0,0 +1,35 @@
...
+      uses: martinbeentjes/npm-get-version-action@v1.3.1
+    - name: ci
+      run: npm ci
+    - name: test
+      run: npm test
+    - name: package
+      run: npm run package
+    - name: rename
+      run: mv vscode-clangd-${{ steps.version.outputs.current-version }}.vsix
+              vscode-clangd-unstable-${{ steps.version.outputs.current-version }}.vsix
```
hokein, line 28: If we're going to make a frequent unstable release (like per commit), I think it would be clearer to use the commit in the name, rather than use the current release version in the package.json (not sure this is possible to get the commit in workflow.)

hokein (commented):
I think it is useful to have this unstable release, it helps prevent rolling out a bad release to users.

.github/workflows/unstable-release.yml
HighCommander4: Makes sense. I updated the action to do this.

hokein (approved):
looks good from my side.

PR #749, opened 2024-12-09
Automatically restart clangd language server after it is installed

k0zmo:


src/install.ts
```diff
@@ -129,6 +132,11 @@ class UI {
     return p;
   }
   set clangdPath(p: string) {
-    config.update('path', p, vscode.ConfigurationTarget.Global);
+    this._pathUpdated =
```
HighCommander4, line 135: I believe a more conventional way to write this is:

```js
this._pathUpdated = new Promise(resolve => {
  config.update('path', ...).then(resolve);
});
```

or alternatively:

```js
this._pathUpdated = new Promise(async resolve => {
  await config.update('path', ...);
  resolve();
});
```

This avoids the extra variable `_resolvePathUpdated`, and also ensures that if the setter is called multiple times, awaiting `_pathUpdated` will correctly wait until the most recent promise is resolved.

src/install.ts
```diff
@@ -81,9 +81,12 @@ class UI {
     }
   }
 
+  private _pathUpdated: Promise<void>|undefined;
+  private _resolvePathUpdated: (() => void)|undefined;
+
   async promptReload(message: string) {
-    if (await vscode.window.showInformationMessage(message, 'Reload window'))
-      vscode.commands.executeCommand('workbench.action.reloadWindow');
+    await this._pathUpdated
```
HighCommander4, line 88: We should null out `_pathUpdated` after having awaited it, to avoid any objects being unnecessarily kept alive

src/install.ts
```diff
@@ -81,9 +81,12 @@ class UI {
     }
   }
 
+  private _pathUpdated: Promise<void>|undefined;
+  private _resolvePathUpdated: (() => void)|undefined;
+
   async promptReload(message: string) {
```
HighCommander4, line 87: I don't think we should ignore `message` entirely. Can we still show it in an info dialog, without waiting for the user to interact with it?

HighCommander4 (changes requested):
Thanks for the patch. The general approach looks reasonable to me. A few comments on the specifics:

src/install.ts
k0zmo: I went with first option as `config.update` is not awaitable.

src/install.ts
k0zmo: I think we can do this:
```ts
  async promptReload(message: string) {
    vscode.window.showInformationMessage(message)
    await this._pathUpdated;
    this._pathUpdated = null;
    vscode.commands.executeCommand('clangd.restart');
  }
```
Without `await`, this won't wait for user prompt and it would serve as a status message `'clangd ${version} is installed'`. Is this what you had in mind?

HighCommander4: Yep, exactly

HighCommander4 (approved):
LGTM

PR #758, opened 2024-12-16
In VsCode 1.72 Node was updated to version 20, before this upgrade, it was possible to execute cmd/bat scripts as executable.

JVApen:
After this update this was suddenly broken.

Fixes clangd#683

In this reapply, we quote both the command line and the arguments, such that both can contain spaces. We also introduce the option useScriptAsExecutable, such that this is only enabled when the user wants it. We also provide the default value as boolean instead of string

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -48,7 +48,7 @@ recommend to run `npm run format` before sending a patch.
 
 To create a new release, create a commit that:
 
-- increases the version number in `package.json`
+- increases the version number in `package.json` and `package-lock.json`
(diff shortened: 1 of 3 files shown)


HighCommander4:
> We also provide the default value as boolean instead of string

Ah, that's why the default value [wasn't working](https://github.com/clangd/vscode-clangd/issues/683#issuecomment-2493044295)!

HighCommander4 (approved):

HighCommander4:
Posted a build including this change at https://github.com/clangd/vscode-clangd/releases/tag/snapshot_20241222

PR #769, opened 2025-01-25
Handle workspace symbols searches with a '::' prefix

joelhock:
There is some prior discussion in #31 that doesn't directly address this issue--it appears to me searching for a symbol with a leading `::` always fails: results from clangd strip the leading `::`, so vscode fuzzy match will filter out all results unless we add prefix back, which is easy enough to do.


HighCommander4 (approved):
Thanks! This looks like a nice improvement to me.

PR #796, opened 2025-04-05
Support commands in config

### LINKED ISSUE #48: support commands in settings
For proper CMake support we would need `"clangd.arguments": [ "--compile-commands-dir=${command:cmake.launchTargetDirectory}" ]`.
https://github.com/microsoft/vscode-cmake-tools/issues/654

Problem is that invoking commands is async so you can't just add this in config.ts:
```js
const cmdPrefix = 'command:';
if (name.startsWith(cmdPrefix))
    return await vscode.commands.executeCommand(name.substr(cmdPrefix.length));
```

Btw, looks like the relevant vscode feature request is https://github.com/microsoft/vscode/issues/46471.

nisargjhaveri:
Fix #48.

- Make `config.get` async, change all usage to support this.
- Add a new replacement for `command:` variables.

### DIFF
--- a/src/config-file-watcher.ts
+++ b/src/config-file-watcher.ts
@@ -19,9 +19,9 @@ class ConfigFileWatcherFeature implements vscodelc.StaticFeature {
   constructor(private context: ClangdContext) {}
   fillClientCapabilities(capabilities: vscodelc.ClientCapabilities) {}
 
-  initialize(capabilities: vscodelc.ServerCapabilities,
+  async initialize(capabilities: vscodelc.ServerCapabilities,
              _documentSelector: vscodelc.DocumentSelector|undefined) {
-    if (!config.get<boolean>('onConfigChangedForceEnable') &&
+    if (!await config.get<boolean>('onConfigChangedForceEnable') &&
         (capabilities as ClangdClientCapabilities)
             .compilationDatabase?.automaticReload) {
       return;
--- a/src/config-file-watcher.ts
+++ b/src/config-file-watcher.ts
@@ -4,8 +4,8 @@ import * as vscodelc from 'vscode-languageclient/node';
 import {ClangdContext} from './clangd-context';
 import * as config from './config';
 
-export function activate(context: ClangdContext) {
-  if (config.get<string>('onConfigChanged') !== 'ignore') {
+export async function activate(context: ClangdContext) {
+  if (await config.get<string>('onConfigChanged') !== 'ignore') {
     context.client.registerFeature(new ConfigFileWatcherFeature(context));
   }
 }
@@ -82,7 +82,7 @@ class ConfigFileWatcher implements vscode.Disposable {
     if ((await vscode.workspace.fs.stat(uri)).size <= 0)
       return;
 
-    switch (config.get<string>('onConfigChanged')) {
+    switch (await config.get<string>('onConfigChanged')) {
     case 'restart':
       vscode.commands.executeCommand('clangd.restart');
       break;
--- a/src/config-file-watcher.ts
+++ b/src/config-file-watcher.ts
@@ -20,7 +20,7 @@ class ConfigFileWatcherFeature implements vscodelc.StaticFeature {
   fillClientCapabilities(capabilities: vscodelc.ClientCapabilities) {}
 
   async initialize(capabilities: vscodelc.ServerCapabilities,
-             _documentSelector: vscodelc.DocumentSelector|undefined) {
+                   _documentSelector: vscodelc.DocumentSelector|undefined) {
     if (!await config.get<boolean>('onConfigChangedForceEnable') &&
         (capabilities as ClangdClientCapabilities)
             .compilationDatabase?.automaticReload) {
--- a/src/config.ts
+++ b/src/config.ts
@@ -71,10 +71,14 @@ async function replacement(name: string): Promise<string|undefined> {
   }
   const commandPrefix = 'command:';
   if (name.startsWith(commandPrefix)) {
+    const commandId = name.substr(commandPrefix.length);
     try {
-      return await vscode.commands.executeCommand(
-          name.substr(commandPrefix.length));
-    } catch {
+      return await vscode.commands.executeCommand(commandId);
+    } catch (error) {
+      console.warn(`Clangd: Error resolving command '${commandId}':`, error);
+      vscode.window.showWarningMessage(
+          `Clangd: Failed to resolve ${commandId}`);
+
       return undefined;
     }
   }
--- a/src/config.ts
+++ b/src/config.ts
@@ -3,8 +3,8 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 
 // Gets the config value `clangd.<key>`. Applies ${variable} substitutions.
-export function get<T>(key: string): T {
-  return substitute(vscode.workspace.getConfiguration('clangd').get<T>(key)!);
+export async function get<T>(key: string): Promise<T> {
+  return await substitute(vscode.workspace.getConfiguration('clangd').get<T>(key)!);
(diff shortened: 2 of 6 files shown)


frayien:
Hi, I tried your branch and it did not work, I got [objet Promise] substitution instead of the actual value ...

Here is a patch to add missing awaits
[missing-await.patch](https://github.com/user-attachments/files/19939967/missing-await.patch)

Got it to work with this, can confirm it is much better than any workaround, thanks you for your PR ! Hope it'll get merged :)

nisargjhaveri:
@frayien, thanks a lot for testing out and suggesting the fix. My bad for missing these cases. I've updated the PR, please have a look?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## Version 0.2.1: June 2, 2025
+
+* Support commands in config [#796](https://github.com/clangd/vscode-clangd/pull/796) [#48](https://github.com/clangd/vscode-clangd/issues/48)
```
HighCommander4, line 5: If you're going to do a new release as part of this PR, please list other new changes since the last release:

 * #822
 * #837
 * #848

HighCommander4 (commented):
@kadircet, do you have any concerns related to security or performance about invoking arbitrary vscode commands from `config.get`?

Otherwise, this looks fine to me on a high level. I haven't looked at the implementation in detail yet, will try to do so in the coming days.

src/config.ts
```diff
@@ -14,19 +14,23 @@ export function update<T>(key: string, value: T,
...
-function substitute<T>(val: T): T {
+async function substitute<T>(val: T): Promise<T> {
   if (typeof val === 'string') {
-    val = val.replace(/\$\{(.*?)\}/g, (match, name) => {
-      // If there's no replacement available, keep the placeholder.
-      return replacement(name) ?? match;
-    }) as unknown as T;
+    const replacementPattern = /\$\{(.*?)\}/g;
+    const replacementPromises: Promise<string | undefined>[] = [];
+    val.replace(replacementPattern, (match, name) => {
```
HighCommander4, line 21: This can be done a bit more efficiently, since the first time around we need only match, not replace:

```ts
const matches = val.matchAll(replacementPattern);
for (const match of matches) {
  // match[1] is the first captured group
  replacementPromises.push(replacement(match[1]));
}
```

src/config.ts
```diff
@@ -14,19 +14,23 @@ export function update<T>(key: string, value: T,
...
+      replacementPromises.push(replacement(name));
+      return match;
+    });
+    const replacements = await Promise.all(replacementPromises);
+    val = val.replace(/\$\{(.*?)\}/g, (match, _) => replacements.shift() ?? match) as unknown as T;
```
HighCommander4, line 26: * Use `replacementPattern` rather than repeating the regex
* Preserve the comment `// If there's no replacement available, keep the placeholder.`, which explains why we do the `?? match` part.
* nit: the function can be written a bit more simply as`match => replacements.shift() ?? match`

So, putting these together, suggested rewrite:

```ts
val = val.replace(replacementPattern,
    // If there's no replacement available, keep the placeholder.
    match => replacements.shift() ?? match) as unknown as T;
```

src/inactive-regions.ts
```diff
@@ -45,18 +45,18 @@ export class InactiveRegionsFeature implements vscodelc.StaticFeature {
       };
     }
   }
-  initialize(capabilities: vscodelc.ServerCapabilities,
+  async initialize(capabilities: vscodelc.ServerCapabilities,
```
HighCommander4, line 48: It's not clear to make whether we can make `initialize` async.

The framework is going to [call it](https://github.com/microsoft/vscode-languageserver-node/blob/d90a87f9557a0df9142cfb33e251cfa6fe27d970/client/src/common/client.ts#L2101) without using `await`... but on the other hand, I don't think it expects any particular side effect to have occurred straight away (unlike, say, `fillClientCapabilities`).

@dbaeumer, could you give us advice on this: will anything bad happen if we make a `StaticFeature` implementation's `initialize` method async?

src/config-file-watcher.ts
```diff
@@ -4,8 +4,8 @@ import * as vscodelc from 'vscode-languageclient/node';
 import {ClangdContext} from './clangd-context';
 import * as config from './config';
 
-export function activate(context: ClangdContext) {
-  if (config.get<string>('onConfigChanged') !== 'ignore') {
+export async function activate(context: ClangdContext) {
```
HighCommander4, line 7: The [call site](https://github.com/clangd/vscode-clangd/blob/2b21206d38c5dab72f7f9fd911b72e4827027e05/src/clangd-context.ts#L215) of this needs to be modified to use `await`.

HighCommander4 (changes requested):
Ok, I looked through the implementation and left some comments.

There's an open question on whether this approach is viable (see the comment about making `initialize` async), let's wait for feedback on that.

HighCommander4:
From a use case point of view, is there anyone looking to put commands into random config values like `"clangd.checkUpdates"`? Could we make the implementation simpler without losing functionality if we allowed them only in `"clangd.arguments"`?

CHANGELOG.md
HighCommander4: Yes, good catch. Still getting the hang of the new version numbering system.

HighCommander4: Works for me.

nisargjhaveri:
> 
From a use case point of view, is there anyone looking to put commands into random config values like "clangd.checkUpdates"? Could we make the implementation simpler without losing functionality if we allowed them only in "clangd.arguments"?

I can see use cases where someone might want to run commands (either custom or provided by other extensions) in configs like `clangd.path` or even `clangd.enable`.

We can explore selectively enabling, not sure thought if that will make it simpler though.

src/config-file-watcher.ts
nisargjhaveri: Fixed now. I thought it would be harmless to call registerFeature later, and it likely would be in this case, does not seem to be the expectation.

On the similar note, we probably want to move `switchSourceHeader.activate(this);` which also calls into registerFeature at the call site before `this.client.start();`

src/inactive-regions.ts
nisargjhaveri: Had a cursory look, as you mentioned I don't think this is expected to have any immediate side-effects. Should be okay.

src/config.ts
nisargjhaveri: Updated

HighCommander4 (approved):
LGTM

Let's wait a couple of days in case @kadircet or @dbaeumer have any thoughts, otherwise this should be good to merge.

src/config.ts
```diff
@@ -60,6 +69,15 @@ function replacement(name: string): string|undefined {
...
     return (typeof config === 'string') ? config : undefined;
   }
+  const commandPrefix = 'command:';
+  if (name.startsWith(commandPrefix)) {
+    try {
+      return await vscode.commands.executeCommand(
+          name.substr(commandPrefix.length));
+    } catch {
+      return undefined;
+    }
```
marcogmaia, line 79: What about showing the user a message that the command couldn't be resolved?

```ts
      console.warn(`Clangd: Error resolving command '${commandId}':`, error);
      vscode.window.showWarningMessage(`Clangd: Failed to resolve ${command}`);
```

nisargjhaveri: @HighCommander4, is this something we'd desire? Currently, we'd just keep the original command variable as is  silently if not resolved.

CHANGELOG.md
nisargjhaveri: Created #856 for the new pre-release.

Issue #498, opened 2023-07-05
Improve multi-project support in a single workspace


tortmayr:
For many C/C++ based Editors it is a common practice to work on several projects within the same workspace simultaneously (e.g. Eclipse CDT). Unfortunately this usecase is currently not well supported when using (vscode-)clangd.

Clangd has no explicit support for mulit-root workspaces. It is possible to provide baseline multi-project support by setting up project-specific compilation databases, since clangd searches and uses the individual compile_commands.json file in the source directory path to determine the CBD. However, this approach has several drawbacks:
Indexing problems occurs when functions with same signature has been defined in separate projects (see this [Issue #38](https://github.com/clangd/vscode-clangd/issues/38#issuecomment-1074670744))
The indexing problems also result in incorrect/unexpected results when using LSP features like `go to definition`
- https://github.com/clangd/clangd/issues/1258
- https://github.com/clangd/clangd/issues/1400
- 
Determining the CDB for external header files comes into troubles when there are several project roots. (see discussion in this [Issue #907](https://github.com/clangd/clangd/discussions/907#discussioncomment-1539520))

In addition, this approach causes issues with functions that should be scoped to a specific project like “finding all references” or “Renaming” (see also [this issue](https://github.com/clangd/vscode-clangd/issues/455))

Other parties like the cdt-lsp team also have encountered [similar problems](https://github.com/clangd/clangd/issues/1549).

We worked on a [proof of concept implementation](https://github.com/eclipsesource/vscode-clangd/tree/multi-client) that solves the issues mentioned above by extending vscode-clangd to support multiple clangd servers within the same workspace: For each project we start a dedicated clangd server and map all files that can be matched to a specific project to its respective clangd server instance. This ensures that files within a project are correctly scoped within the boundaries of a single project, resolving all issues mentioned above.
In addition, we track the currently active project. Tracking the currently active project (or rather its clangd server) ensures that extensions that want to communicate with the clangd server without a specific file context (e.g. the memory usage view) still work as expected. The active project can either be set manually by the user, or can be tracked automatically based on the last active file editor.

This multi-project feature is completely opt–in and can be activated via VS Code preferences. The default project resolution strategy is to map each workspace folder to one project. An extension mechanism is provided to support more complex project resolution strategies.

[...]

PR #810, opened 2025-05-18
Add support for multi-root workspaces

### LINKED ISSUE #38: Support for multi-root workspaces
As multi-root workspaces become more popular and widely adopted, it would be great if `vscode-clangd` could support this as well.

There are many use-cases for C/C++ projects where this would be useful (working with different hardware targets, working across third party libraries, etc.). I imagine there would need to be a fair amount of discussion on how to approach this. Some items for debate are whether the plugin should spawn an instance of clangd per root or whether it should be allowed for the client to [specify multiple compilation databases](https://github.com/clangd/clangd/issues/116).

Some relevant links to study on the LSP side are the [vscode LSP multi-server example](https://github.com/microsoft/vscode-extension-samples/tree/master/lsp-multi-server-sample) which illustrates how to start a server per wokspace folder, and the [LSP specification for handling workspace folders](https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders). Feel free to discuss below any ideas for how to implement this!

DaanDeMeyer:
These changes add proper support for multi root workspaces with a clangd server per workspace folder.

On a high level, we have a single language client per folder, and have to rework things so that various global objects (views, ...) are actually global and not instantiated per client anymore.

The important use case this enables is where clangd is invoked within a container per process with all the dependencies installed in the container. To make this use case work in a multi-root setup, a clangd instance per workspace folder has to be spawned.

Fixes #38

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,14 +1,18 @@
-import {BaseLanguageClient} from 'vscode-languageclient';
-
-import {ClangdApiV1, ClangdExtension} from '../api/vscode-clangd';
+import {ClangdApiV1, ClangdApiV2, ClangdExtension} from '../api/vscode-clangd';
+import {ClangdContext} from './clangd-context';
 
 export class ClangdExtensionImpl implements ClangdExtension {
-  constructor(public client: BaseLanguageClient|undefined) {}
+  constructor(public context: ClangdContext) {}
 
   public getApi(version: 1): ClangdApiV1;
+  public getApi(version: 2): ClangdApiV2;
   public getApi(version: number): unknown {
     if (version === 1) {
-      return {languageClient: this.client};
+      return {languageClient: this.context.clients.entries().next().value?.[1]};
+    }
+
+    if (version === 2) {
+      return {languageClients: [...this.context.clients.values()]};
     }
 
     throw new Error(`No API version ${version} found`);
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -3,61 +3,17 @@
 import * as vscode from 'vscode';
 import * as vscodelc from 'vscode-languageclient/node';
 
-import {ClangdContext} from './clangd-context';
+import {ClangdContext, ClangdLanguageClient} from './clangd-context';
 import type {ASTParams, ASTNode} from '../api/vscode-clangd';
 
 const ASTRequestMethod = 'textDocument/ast';
 
-export function activate(context: ClangdContext) {
-  const feature = new ASTFeature(context);
-  context.client.registerFeature(feature);
-}
-
 const ASTRequestType =
     new vscodelc.RequestType<ASTParams, ASTNode|null, void>(ASTRequestMethod);
 
-class ASTFeature implements vscodelc.StaticFeature {
-  constructor(private context: ClangdContext) {
-    // The adapter holds the currently inspected node.
-    const adapter = new TreeAdapter();
-    // Create the AST view, showing data from the adapter.
-    const tree =
-        vscode.window.createTreeView('clangd.ast', {treeDataProvider: adapter});
-    context.subscriptions.push(
-        tree,
-        // Ensure the AST view is visible exactly when the adapter has a node.
-        // clangd.ast.hasData controls the view visibility (package.json).
-        adapter.onDidChangeTreeData((_) => {
-          vscode.commands.executeCommand('setContext', 'clangd.ast.hasData',
-                                         adapter.hasRoot());
-          // Work around https://github.com/microsoft/vscode/issues/90005
-          // Show the AST tree even if it's been collapsed or closed.
-          // reveal(root) fails here: "Data tree node not found".
-          if (adapter.hasRoot())
-            // @ts-ignore
-            tree.reveal(null);
-        }),
-        // Create the "Show AST" command for the context menu.
-        // It's only shown if the feature is dynamicaly available (package.json)
-        vscode.commands.registerTextEditorCommand(
-            'clangd.ast',
-            async (editor, _edit) => {
-              const converter = this.context.client.code2ProtocolConverter;
-              const item =
-                  await this.context.client.sendRequest(ASTRequestType, {
-                    textDocument:
-                        converter.asTextDocumentIdentifier(editor.document),
-                    range: converter.asRange(editor.selection),
-                  });
-              if (!item)
-                vscode.window.showInformationMessage(
-                    'No AST node at selection');
-              adapter.setRoot(item ?? undefined, editor.document.uri);
-            }),
-        // Clicking "close" will empty the adapter, which in turn hides the
-        // view.
-        vscode.commands.registerCommand(
-            'clangd.ast.close', () => adapter.setRoot(undefined, undefined)));
+export class ASTFeature implements vscodelc.StaticFeature {
+  constructor(client: ClangdLanguageClient) {
+    client.registerFeature(this);
   }
 
   fillClientCapabilities(capabilities: vscodelc.ClientCapabilities) {}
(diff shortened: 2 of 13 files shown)


DaanDeMeyer:
This currently doesn't work because the second client tries to register `clangd.applyFix` command twice which fails. This is done by the clangd server so it's not something I can fix in this PR. 

https://github.com/microsoft/vscode-languageserver-node/issues/607 and https://github.com/microsoft/vscode-languageserver-node/issues/333 discuss this issue with https://github.com/microsoft/vscode-languageserver-node/issues/333 suggesting to use random UUIDs instead of hardcoded command names but I'm not sure how this would work for clangd.

src/api.ts
```diff
@@ -1,14 +1,18 @@
...
 export class ClangdExtensionImpl implements ClangdExtension {
-  constructor(public client: BaseLanguageClient|undefined) {}
+  constructor(public context: ClangdContext) {}
 
   public getApi(version: 1): ClangdApiV1;
+  public getApi(version: 2): ClangdApiV2;
   public getApi(version: number): unknown {
     if (version === 1) {
-      return {languageClient: this.client};
+      return {languageClient: this.context.clients.entries().next().value?.[1]};
```
JVApen, line 11: Should we print out a warning when API1 is used in combination with multiple roots?

src/ast.ts
```diff
@@ -108,10 +64,55 @@ function describe(role: string, kind: string): string {
 }
 
 // Map a root ASTNode onto a VSCode tree.
-class TreeAdapter implements vscode.TreeDataProvider<ASTNode> {
+export class ASTProvider implements vscode.TreeDataProvider<ASTNode> {
```
JVApen, line 67: What do these changes have to do with Multi-Root? I seem to be missing some context

src/clangd-context.ts
```diff
@@ -199,39 +235,82 @@ export class ClangdContext implements vscode.Disposable {
...
+            'textDocument/clangd.fileStatus',
+            (fileStatus) => { this.fileStatus.onFileUpdated(fileStatus); });
+      } else if (newState === vscodelc.State.Stopped) {
+        // Clear all cached statuses when clangd crashes.
+        this.fileStatus.clear();
+      }
+    });
+
+    client.start();
     console.log('Clang Language Server is now active!');
```
JVApen, line 266: Can you update this message to include the working directory? That way, it is clear to users that are less familiar with the internals of the extension why multiple LSPs are active.

src/extension.ts
```diff
@@ -39,34 +47,51 @@ export async function activate(context: vscode.ExtensionContext):
           return;
         }
 
+        const client = clangdContext.getActiveClient();
+        if (client === undefined) {
```
JVApen, line 51: If your clangd crashed (or was explicitly killed by the user), the restart command used to start it again. Is this still the case?

src/extension.ts
```diff
@@ -39,34 +47,51 @@ export async function activate(context: vscode.ExtensionContext):
...
+        clangdContext.addClient(folder);
       }));
 
+  function didOpenTextDocument(document: vscode.TextDocument) {
+    if (!isClangdDocument(document) || document.uri.scheme !== 'file') {
+      return;
+    }
+
+    let folder = vscode.workspace.getWorkspaceFolder(document.uri);
+    if (folder === undefined) {
```
JVApen, line 75: If you are in your file and do a go-to include of the standard library, will clangd still work on those headers?

src/memory-usage.ts
```diff
@@ -48,39 +43,43 @@ function convert(m: WireTree, title: string): InternalTree {
...
-                                   'memoryUsageProvider' in capabilities);
+    if ('memoryUsageProvider' in capabilities)
+      vscode.commands.executeCommand('setContext', 'clangd.memoryUsage.supported', true);
   }
   getState(): vscodelc.FeatureState { return {kind: 'static'}; }
   clear() {}
 }
 
-class TreeAdapter implements vscode.TreeDataProvider<InternalTree> {
+export class MemoryUsageProvider implements vscode.TreeDataProvider<InternalTree> {
```
JVApen, line 63: I'm wondering if making this feature workspace specific is the right thing to do. My first reaction would be to expect a tree node per worktree, such that you can get a global overview. What do you think?

JVApen (changes requested):

JVApen:
Given the impact of this change, can you change the version number to 0.2.1 (or another patch number if that's already used) such that it triggers a pre-release version?

DaanDeMeyer:
It does not work yet. Changes have to be made to clangd to make this possible at all.

DaanDeMeyer:
> Does having a multi-root workspace need to imply using a separate clangd instance for each root? I can see use cases where this is desirable, but also other use cases where having the different roots share a single instance is desirable (e.g. to allow navigation between a library and an application that uses it).

It does not, I reworked this to make the behavior configurable with a new `enablePerFolderServer` option.

> Are you aware of the previous work on this referenced in https://github.com/clangd/vscode-clangd/issues/498?

I am not, this was developed completely separately.

src/extension.ts
DaanDeMeyer: I will need to test this in more detail but I expect not. For my use case the standard library headers will be part of the workspace folder so that is easy to handle but when they live outside of the workspace folder I'm not sure how to handle.

package.json
```diff
@@ -95,13 +95,13 @@
                 "clangd.path": {
                     "type": "string",
                     "default": "clangd",
-                    "scope": "machine-overridable",
```
JVApen: For my understanding, what impact does this change have?

DaanDeMeyer: These will now be synced with settings sync. I'm actually unsure if this change is required now that I look at it again, I'll probably revert it.

DaanDeMeyer:
> > @JVApen Ping @HighCommander4 Ping, any way I can help with getting this reviewed?
> 
> (As always, I can only speak for myself.)
> 
> Given that I'm involved in clangd-related projects on a purely volunteer basis with no employer support, I'm unlikely to have time to spend on reviewing a change of this size in the near future.
> 
> I've pointed you earlier to previous work on this in https://github.com/eclipse-cdt-cloud/vscode-clangd. Have you had a chance to look at that since? Does that project by chance serve your needs? Or could the changes being proposed here, living in a "downstream" plugin similar to that one, serve your needs?

I would very much like these changes to be accepted here rather than in a downstream fork. For the very selfish reasons that I want to be able to use both the extension and clangd itself without having to compile anything from source and having the code upstreamed makes that possible.

As a maintainer myself I'm unfortunately well aware that time is extremely limited and that it helps sometimes to ping once in a while to make sure maintainers don't forget the PR exists. Of course there's no obligation from anyone to look at this, I'm just trying my best to get this merged as it would make my life significantly easier.

HighCommander4:
> I want to be able to use both the extension and clangd itself without having to compile anything from source

Why does using a different extension (e.g. a hypothetical vscode-clangd-multi-project extension rather than the vscode-clangd extension) require compiling anything from source?

src/api.ts
```diff
@@ -1,16 +1,54 @@
...
 
   public getApi(version: 1): ClangdApiV1;
+  public getApi(version: 2): ClangdApiV2;
   public getApi(version: number): unknown {
     if (version === 1) {
-      return {languageClient: this.client};
+      // V1 backward compatibility: return the active context's client
+      const manager = this.manager;
+      return {
+        get languageClient() { return manager.getActiveContext()?.client; },
```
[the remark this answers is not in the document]
DaanDeMeyer: @JVApen So previously, we'd create a clangdcontext when the extension was activated, regardless if any c/c++ or other file type managed by clangd was open or not. On the other hand, the activation events listed in package.json are the following:

```
    "activationEvents": [
        "onLanguage:c",
        "onLanguage:cpp",
        "onLanguage:cuda-cpp",
        "onLanguage:objective-c",
        "onLanguage:objective-cpp"
    ],
```

so I think previously the extension was only activated when a relevant file was opened in the first place.

With these changes, when the extension is activated, we'll still only start a context when a file of one of the activation languages is opened. But given the extension is only activated when a file of one of the activation languages is opened, I think in practice nothing changes, and there will always be a context to return a client from.

To summarize, I think V1 won't return null in more cases than before.

Issue #765, opened 2025-01-08
Publish pre-release extensions in the store


JVApen:
Recently the creation of pre-release versions of the extension was added. This publishes to GitHub.

Visual Studio Code has the concept of pre-release built-in, so it would be useful to make use of that. As such, we would get a larger reach and reduce the burden for people to test this.

Documentation about this can be found [here](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#:~:text=Users%20can%20install%20pre%2Drelease,before%20the%20official%20extension%20release.&text=We%20only%20support%20major.,release%20tags%20are%20not%20supported.)

HighCommander4:
I'm trying to think through the logistics of how this should work.

Currently, we have two CI workflows for creating releases:
 * [The one for publishing a stable release](https://github.com/clangd/vscode-clangd/blob/master/.github/workflows/auto-release.yml). This runs automatically on every commit, checks whether the commit updated the version field in `package.json`, and if so publishes a new stable release to Github, the VSCode Marketplace, and OpenVSX.
 * [The one for publishing an unstable release](https://github.com/clangd/vscode-clangd/blob/master/.github/workflows/unstable-release.yml), newly added in https://github.com/clangd/vscode-clangd/pull/748. This is manually triggered, and published a new pre-release to Github only. Unlike the stable releases, the release name (and package name) contain a commit SHA rather than a version number.

If we're going to publish unstable releases to the VSCode Marketplace, I guess we'll want to identify them with a version number rather than a SHA? And maybe it doesn't need to be a dedicated version number, maybe we can "promote" an unstable release package itself to stable?

One potential way it could work is:
 * Change the automatic workflow to create an **unstable** release (pre-release) when the version number is bumped. Now, _every_ new version starts as unstable.
 * Have a manually triggered workflow that promotes the latest unstable release to stable?

I'm open to suggestions / alternatives.

JVApen:
I see 2 flows here:
 - either release based on an existing pre-release when a newer pre-release was created
 - only allow releasing the last pre-release

I'd be inclined to use the second flow as it reduces complexity while releasing an older pre-release seems rather exceptional. (A revert can be used in that case)

As such, I think that:
- the release notes should contain a new header for the pre-release
- each PR should extend the release notes (if relevant)
- each PR results in an automatic pre-release build (bumps the patch version automatically?)
- releasing requires a special commit (like today) that renames the pre-release header in the release notes and bumps the major version number
[...]

PR #817, opened 2025-06-02
Release 0.2.0

HighCommander4:


HighCommander4:
Note that, since https://github.com/clangd/vscode-clangd/pull/777, only versions with an even minor version number are published as stable, so I'm bumping the minor version to 0.2 to release the changes contained in versions 0.1.34 and 0.1.35 (which have only been published as pre-releases) as stable.

HighCommander4:
@kadircet @hokein Also just wanted to kindly repeat my request to add @JVApen as a project member (or alternatively make me an admin so I can add new project members), who indicated he would be willing to help with reviews.

JVApen:
LGTM

JVApen (approved):

kadircet:
thanks for mentioning this, i definitely didn't notice that especially as https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd displays latest release (without any pre-release tags/indications).

there wasn't much context in #777 but why do we need this exactly? especially at the cost of some version numbers. can we at least make this based on the `patch` version, rather than the minor component? it feels surprising that anything we push from now on `(0.2.x)` will actually become "stable" immediately.

how's this going to look like in practice going forward? do we have a `0.3.X` release first (because we want to take precedence over `0.2.Y` stable channel), let people test it, and release to `0.2.Y` afterwards as stable?

JVApen:
@kadircet The suggestion to use even and odd numbers for release and pre-release come from the vs code documentation.
I agree that your proposal might be more initiative, though the current system was easier to implement.

HighCommander4:
> there wasn't much context in #777 but why do we need this exactly? especially at the cost of some version numbers. can we at least make this based on the `patch` version, rather than the minor component? it feels surprising that anything we push from now on `(0.2.x)` will actually become "stable" immediately.

Github didn't do a good job of linking to the issue referenced in the PR title, https://github.com/clangd/vscode-clangd/issues/765, but there is some additional context there.

The guidance on the usage of version numbers in particular comes from [this VSCode documentation page](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions):

> VS Code will automatically update extensions to the highest version available, so even if a user opted-into a pre-release version and there is an extension release with a higher version, the user will be updated to the released version. So, we recommend that extensions use `major.EVEN_NUMBER.patch` for release versions and `major.ODD_NUMBER.patch` for pre-release versions. For example: `0.2.*` for release and `0.3.*` for pre-release.

So my understanding is that, if we follow this guidance, the next unstable version we publish would be 0.3.x, and the next stable version after that, 0.4.x, and so on. That's a bit of a change to how we manage version numbers compared to before (it will be the minor version that will go up into the tens in due course, rather than the patch version), but that seems sensible to me:
 * the minor version is what gets incremented every stable release (to the next even number), with odd numbers in between for unstable releases
 * the patch version can be used to putting out multiple unstable releases between stable releases without using up a minor version number
 * the major version is still available if at some point we want to communicate that we have a stable API and follow semver

PR #822, opened 2025-06-16
Fix onConfigChanged.forceEnable

briankendall:
`onConfigChanged.forceEnable` doesn't work due to a naming conflict with `onConfigChanged`. If the latter is defined then the former can't be set.

This PR simply renames `onConfigChanged.forceEnable` to `onConfigChangedForceEnable`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
                 "clangd.onConfigChanged": {
                     "type": "string",
                     "default": "prompt",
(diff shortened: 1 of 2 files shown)


HighCommander4 (approved):
Thanks for tracking that down! (This actually came up during the testing of the [original patch](https://github.com/clangd/vscode-clangd/pull/688#discussion_r1797492763), but then it got overlooked when the patch was [resubmitted](https://github.com/clangd/vscode-clangd/pull/808).)

Issue #839, opened 2025-07-15
[Feature Request] Add "Create Source/Header Pair" Command


anonchihaya0908:
Hello,

Thank you for your excellent work on this extension.

##### Summary

To improve the C++ development workflow, I propose adding a new command: **Create Source/Header Pair**. This feature would allow users to create a `.h` and `.cpp` file simultaneously from a single class name, similar to the functionality found in IDEs like CLion.

##### The Problem (User Story)

Currently, when starting a new class, a developer has to perform several manual steps:
1.  Right-click in the explorer -> New File -> `MyClass.h`.
2.  Write the header guards and class boilerplate code.
3.  Right-click in the explorer -> New File -> `MyClass.cpp`.
4.  Write the `#include "MyClass.h"` and basic method implementations.

This process is repetitive and slows down the development flow.

##### Proposed Solution

I suggest a new command, `clangd.createSourceHeaderPair`, with the title "Create Source/Header Pair". When executed, it would:

1.  Prompt the user for a class name (e.g., `MyClass`).
2.  **Intelligently determine the target directory** (preferring the directory of the active file, otherwise falling back to the workspace root).
3.  Create two files in that directory:
    *   `MyClass.h` (with a basic header template, including guards and class definition).
    *   `MyClass.cpp` (with the corresponding include and empty constructor/destructor).
4.  Automatically open the newly created header file in the editor for immediate use.

This command would be accessible from both the Command Palette and the editor's context menu.

##### Demo

Here is a short video demonstrating the proposed feature:

![Demo of Create Source/Header Pair](https://github.com/user-attachments/assets/24f249a9-724a-4eaf-9177-fd103b679ae1)

##### Future Enhancements

I have also considered supporting more file extensions (e.g., `.hpp`, `.cc`). This could be implemented in a follow-up PR by making the pairing rules user-configurable in the settings. For now, this initial proposal focuses on the core `.h`/`.cpp` functionality.

##### Contribution

I have already implemented a working version of this feature that aligns with the project's existing coding style and architecture. The code is ready for review.

If this proposal is something you would consider for the extension, I am happy to open a Pull Request immediately.

Thank you for your time and consideration. I look forward to your feedback.

anonchihaya0908:
Just wanted to post a final summary of what this PR currently accomplishes and what I see as potential next steps, based on all the great feedback and discussion in issue #839.

**What this PR delivers:**
*   A new command, "Create Source/Header Pair," accessible from the command palette and context menu.
[...]

PR #840, opened 2025-07-16
feat(command): Add Create Source/Header Pair command

anonchihaya0908:
Improves the C++ development workflow by providing a new command to create corresponding source and header files simultaneously, mirroring common IDE functionality.

This new command, `clangd.createSourceHeaderPair`, provides the following features:
- Prompts the user for a valid C++ class name with robust validation.
- Intelligently detects the target directory, preferring the active file's location.
- Generates professional-grade file templates for both the header and source file.
- Handles cross-platform line endings (EOL) correctly by respecting user settings.
- Prevents accidental file overwrites by checking for existing files before creation.
- Is accessible from the Command Palette and the editor context menu.

Resolves #[839]

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,11 @@
                 "title": "Switch Between Source/Header"
             },
             {
+                "command": "clangd.createSourceHeaderPair",
+                "category": "clangd",
+                "title": "Create Source/Header Pair"
+            },
+            {
                 "command": "clangd.install",
                 "category": "clangd",
                 "title": "Download language server"
@@ -332,6 +337,11 @@
                     "group": "0_navigation@5"
                 },
                 {
+                    "command": "clangd.createSourceHeaderPair",
+                    "when": "resourceLangId == c || resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp",
+                    "group": "1_modification@1"
+                },
+                {
                     "command": "clangd.ast",
                     "when": "(resourceLangId == c || resourceLangId == cpp || resourceLangId == cuda-cpp || resourceLangId == objective-c || resourceLangId == objective-cpp) && clangd.ast.supported"
                 }
@@ -388,17 +398,20 @@
                 {
                     "id": "clangd.typeHierarchyView",
                     "name": "Type Hierarchy",
-                    "when": "clangd.typeHierarchyVisible"
+                    "when": "clangd.typeHierarchyVisible",
+                    "icon": "$(type-hierarchy)"
                 },
                 {
                     "id": "clangd.memoryUsage",
                     "name": "clangd Memory Usage",
-                    "when": "clangd.memoryUsage.hasData"
+                    "when": "clangd.memoryUsage.hasData",
+                    "icon": "$(dashboard)"
                 },
                 {
                     "id": "clangd.ast",
                     "name": "AST",
-                    "when": "clangd.ast.hasData"
+                    "when": "clangd.ast.hasData",
+                    "icon": "$(list-tree)"
                 }
             ]
         }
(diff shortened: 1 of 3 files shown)


anonchihaya0908:
Hi.
Thank you again for your time and for creating this great extension.
After submitting this PR, I had a follow-up thought I wanted to share. I realized that the current file templates I've implemented are C++ specific (using class, etc.) and wouldn't be suitable for pure C projects.
I've been thinking about the best way to handle this, and wanted to get your opinion.
Perhaps we could treat this current PR as a C++-focused first step? If it's in a good state, it could potentially be merged to benefit C++ users sooner.
Then, for C language support, I was thinking we could address it in a separate, follow-up PR. A possible approach might be to make the pairing rules "language-aware," so the extension could provide different templates for C and C++. I'd be happy to create a new issue to discuss this idea further if you think it's a reasonable path forward.
What are your thoughts on this approach? I'm completely open to feedback and ready to make any changes needed.
Thank you!

anonchihaya0908:
Added intelligent sorting logic for options and more precise expressions.
image
Temporarily locked.cpp/.h.

anonchihaya0908:
Just wanted to post a final summary of what this PR currently accomplishes and what I see as potential next steps, based on all the great feedback and discussion in issue #839.

**What this PR delivers:**
*   A new command, "Create Source/Header Pair," accessible from the command palette and context menu.
*   A multi-step wizard UI that allows users to choose the type of content they want (Class, Struct, Empty, etc.).
*   The wizard is context-aware: it intelligently reorders the template choices based on whether the user is in a C or C++ file.
*   It includes robust features like input validation, file existence checks, platform-aware EOL handling, and smart placeholder generation.
*   Crucially, it correctly infers the language context for `.h` files by checking for companion `.c` source files, solving a key usability issue.

**Current Limitations & Future Work:**
*   **Hardcoded Extensions:** As a focused first step, this implementation currently creates only `.h`/.`cpp` or `.h`/.`c` pairs. The logic for user-configurable extensions (like `.hpp` or `.cc`) is something I've planned as a clear follow-up enhancement in a separate PR.
*   **Complex Project Structures:** The context detection relies on finding companion files in the same directory. A future improvement could be to support more complex structures (e.g., `src/` and `include/` folders), potentially by leveraging `compile_commands.json` or other project-level information.

I believe this PR, in its current state, provides a solid and valuable foundation. Thank you again for all the guidance. I'm looking forward to your review!

anonchihaya0908:
https://github.com/anonchihaya0908/vscode-clangd/tree/Create-Configurable-Extension-name

anonchihaya0908:
finished.

package.json
JVApen, line 1: I like your addition of relevant icons, though can you do this in a separate PR, such that it's easier to track the relevant changes and/or land these changes separately?

src/ast.ts
```diff
@@ -4,6 +4,7 @@ import * as vscode from 'vscode';
 import * as vscodelc from 'vscode-languageclient/node';
 
 import {ClangdContext} from './clangd-context';
+
```
JVApen, line 7: File is touched without any significant changes. Can you undo this newline? (Unless required by formatting rules)

src/create-source-header-pair/service.ts
JVApen, line 1: This contains a lot of code trying to find out which rule the user intended. Can't we delegate that to the user? If they add a 'label' as part of the configuration, they can have a rule for classes, structs, free functions or anything custom with the exact same extensions. They can than pick the intended entry from a list.

vscode-clangd-fancy-0.2.1.vsix
JVApen, line 1: Binaries should not be included

package-lock.json
```diff
@@ -1,16 +1,17 @@
 {
-    "name": "vscode-clangd",
-    "version": "0.2.0",
+    "name": "vscode-clangd-fancy",
```
JVApen, line 2: Please don't rename the project

JVApen (changes requested):
See comments

Issue #847, opened 2025-07-27
Specifying `clangd.trace` overrides other clangd launch options (cwd, shell)


Rac75116:
When reading /src/clangd-context.ts, it appears that the existing `clangd.options` object is overwritten when a trace file is specified via the `clangd.trace` config:
```ts
    const clangd: vscodelc.Executable = {
      command: clangdPath,
      args: clangdArguments,
      options: {
        cwd: vscode.workspace.rootPath || process.cwd(),
        shell: useScriptAsExecutable
      }
    };
    const traceFile = config.get<string>('trace');
    if (!!traceFile) {
      const trace = {CLANGD_TRACE: traceFile};
      clangd.options = {env: {...process.env, ...trace}};
    }
```

As a result, clangd launch options such as `cwd` and `shell` are discarded if `clangd.trace` is set.

Is this intentional? If not, it might make sense to preserve existing options and only merge in the environment variables like this:
```ts
clangd.options = {
  ...clangd.options,
  env: {...process.env, ...trace}
};
```

This would ensure that cwd, shell, and any other pre-defined options are retained while the CLANGD_TRACE environment variable is added.

Thanks for your consideration!

**System information**
clangd extension version: 0.2.0

Rac75116:
Sure, I'd be glad to submit a PR.

PR #848, opened 2025-07-28
Fix clangd launch options being overwritten when trace file is set

Rac75116:
Fix [#847](https://github.com/clangd/vscode-clangd/issues/847).


HighCommander4 (approved):
Thanks!

PR #856, opened 2025-08-19
Release 0.3.0

nisargjhaveri:
Create new pre-release version to stabilise #796.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## Version 0.3.0: August 19, 2025
+
+* Rename onConfigChanged.forceEnable to onConfigChangedForceEnable [#822](https://github.com/clangd/vscode-clangd/pull/822).
+* Fix UI.localize function to use placeholders [#837](https://github.com/clangd/vscode-clangd/pull/837).
+* Fix clangd launch options being overwritten when trace file is set [#848](https://github.com/clangd/vscode-clangd/pull/848).
+* Support command variables in config [#796](https://github.com/clangd/vscode-clangd/pull/796).
+
 ## Version 0.2.0: June 1, 2025
 
 * Stable release of changes in versions 0.1.34 and 0.1.35.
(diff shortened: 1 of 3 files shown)


HighCommander4 (approved):
LGTM

nisargjhaveri:
@HighCommander4, thanks for merging. Though I don't think it published the pre-release in the [actions](https://github.com/clangd/vscode-clangd/actions/runs/17089853132/job/48461189647). What is the expectation here? Is anything else required to create a release?

JVApen:
I was looking to test this morning and noticed the same, I'll try and take a look later this week

JVApen:
What I already found:
The `detect version` step outputs:
````
Run EndBug/version-check@v2
  with:
    diff-search: true
    file-name: package.json
    token: ***
    github-api-url: https://api.github.com
  env:
    DISPLAY: :99
Searching for version update...
Searching in 1 commit...
  Package file name: "package.json"
  Package file URL: undefined
  Version assumptions: undefined
  - 91427e5: match found, more info below
Found match for version 0.3.0: 91427e5 Release 0.3.0 (#856)
Outputs:
  version: 0.3.0
  type: minor
  commit: 91427e58cd0806e4b4b4c93cb3653fec9c28b313
  changed: false
````

According to the documentation, the output should only contain a version/type/commit if a change was detected, see  https://github.com/EndBug/version-check?tab=readme-ov-file#outputs

Based on this output, I'm assuming:
 - it found the 1 commit that landed and needed to be checked 
 - it found the right file which contains the change
 - it seems to have detected the right type of change and read the version number correctly 
 - yet changed is on false

As the changed is false, no other steps are executed since all conditions depend on this field.

HighCommander4:
@tamird could this be a regression from https://github.com/clangd/vscode-clangd/pull/824 ?

tamird:
Yes, one that appears to have been [fixed yesterday](https://github.com/EndBug/version-check/issues/315)

tamird:
Should be automatic yeah

nisargjhaveri:
Created #858.

PR #875, opened 2025-09-28
Type hierarchy view improvements

stephan-razin:
- Fix collapsible state for classes without children.
- Fix base-classes hierarchy displaying for classes without children.
- Improve icons and details information like native vscode Peek Type Hierarchy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -248,13 +248,13 @@
                 "command": "clangd.typeHierarchy.viewParents",
                 "category": "clangd",
                 "title": "Types: Show Base Classes",
-                "icon": "$(triangle-up)"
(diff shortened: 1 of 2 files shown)


hokein:
> Yeah, I think that, since the only reason we had our own type hierarchy view to begin with was that there wasn't a standard one at the time, we should be moving in the direction of deprecating/removing ours rather than enhancing it.

+1, the existing `clangd-vscode` view should be considered deprecated in favor of the standard one. While I personally find our view which shows both base and derived classes more useful, the extra maintenance burden of a separate hierarchy view is not warranted.

Issue #892, opened 2025-11-12
How to specify platform-specific clangd arguments like `--malloc-trim`?


emmenlau:
I have my clangd arguments in the vscode workspace settings `MyProject.code-workspace`. This helps so that all developers benefit from a good setup directly after cloning the project.

On Linux we urgently require the argument `--malloc-trim` to avoid frequent machine crashes due to out-of-memory. With this option, the RAM usage drops by 50-80%.

However, this option does not exist on Windows. Windows developers get a clangd startup error about the unsupported option.

Is there any way to set this option platform-specific? Or some workaround, to get this going?

HighCommander4:
> Is there any way to set this option platform-specific?

Looks like this is a missing feature in vscode: https://github.com/microsoft/vscode/issues/5595

> Or some workaround, to get this going?

Maybe with an extension like https://marketplace.visualstudio.com/items?itemName=runarsf.platform-settings.

JVApen:
Workspace wins over remote settings over user settings. 
Checking all these levels in-between seems more complex. Also here you have override semantics. So I'd a platform specific setting at any level wins over the generic one.

I think we can keep the complexity contained in the getter for the settings + a bool (with default value) + the package file that has all settings described

emmenlau:
I just wanted to chime in: Would be awesome if this is worked on! For me it has a very high impact, as `--malloc-trim` is basically the only way how I can use `clangd` on the laptop.

PR #907, opened 2025-12-16
Fixes clangd/vscode-clangd#892

JVApen:
add platform specific overrrides for settings for executable name, arguments for the executable and extra compile flags


JVApen:
Ping @HighCommander4

StMartin81:
It would be really nice if this PR would be merged so that we have the option to set platform specific configurations for clangd.

PR #954, opened 2026-05-04
Fix completion crash when item.range starts using {inserting, replacing}

argothiel:
The LSP spec (since 3.16) allows CompletionItem.range to be either a plain Range or an InsertReplaceRange ({inserting, replacing}). The old code cast item.range directly to vscode.Range; once clangd starts sending the newer form, this will crash (llvm/llvm-project#187623).

Narrow the type before extracting the start position, and guard against a missing range entirely.

### DIFF
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -148,7 +148,7 @@ export class ClangdContext implements vscode.Disposable {
               prefix = document.getText(new vscode.Range(start, position));
             }
             if (prefix)
-            item.filterText = prefix + '_' + item.filterText;
+              item.filterText = prefix + '_' + item.filterText;
             // Workaround for https://github.com/clangd/vscode-clangd/issues/357
             // clangd's used of commit-characters was well-intentioned, but
             // overall UX is poor. Due to vscode-languageclient bugs, we didn't
--- a/src/clangd-context.ts
+++ b/src/clangd-context.ts
@@ -138,8 +138,15 @@ export class ClangdContext implements vscode.Disposable {
           let items = (!list ? [] : Array.isArray(list) ? list : list.items);
           items = items.map(item => {
             // Gets the prefix used by VSCode when doing fuzzymatch.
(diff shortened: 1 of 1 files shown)


JVApen (approved):

HighCommander4:
cc @timon-ul

(I'm intending this as a review request, but I can't actually add you to the "Reviewers" field because you're not a member of the clangd org.)

argothiel:
@JVApen Thanks for the review! Could you merge? I don't have permissions.

JVApen:
I wanted to wait until others had the chance to review, in case I missed something. This looks very trivial, though I've learned others have a different way of looking at patches. 

Given the explicit request for @timon-ul, I'd like to wait merging until he responded.

argothiel:
Makes sense, thank you! 🙂

JVApen:
It might make sense to bump the version number here for a new pre-release version and if that's not causing issues, to trigger a real release in a next PR. See https://github.com/clangd/vscode-clangd/pull/876/changes for an example. I would recommend doing this in a separate PR. For X.Y.Z, every odd Y is a pre-release, following the guidelines of vs code

HighCommander4:
(Agree this looks pretty trivial. I just wanted @timon-ul to okay it since it's related to https://github.com/llvm/llvm-project/pull/187623 which he has been reviewing.)

timon-ul:
Just want to say it looks fine, but wanted to be a good reviewer and test the changes, but I am struggeling a bit with setting that up...

timon-ul:
Looks good to me, got it to run, works, code also looks good.

argothiel:
> It might make sense to bump the version number here for a new pre-release version and if that's not causing issues, to trigger a real release in a next PR. See https://github.com/clangd/vscode-clangd/pull/876/changes for an example. I would recommend doing this in a separate PR. For X.Y.Z, every odd Y is a pre-release, following the guidelines of vs code

PR created: #956

PR #961, opened 2026-05-20
Add support for setting OS specific paths for clangd

StMartin81:
Allow setting the paths to the clangd executable depending on the platform. When using the platform specific setting don't use clang.path as this setting will discard all platform specific settings.

This PR was co-authored with AI and tested locally.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -4,9 +4,13 @@
 
 This extension allows configuring command-line flags to the `clangd` executable.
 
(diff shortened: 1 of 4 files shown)


HighCommander4:
I think this may be a subset of https://github.com/clangd/vscode-clangd/pull/907.

StMartin81:
Yes, it looks like this PR is a subset from the other PR.

PR #984, opened 2026-08-01
Implements the client feature of the ASTSearch / clang query capability

Febbe:
- currently mirrors the AST view.
 - requires launching a command via Ctrl + Shift+P (clangd: Search AST).

VSCode implementation for https://github.com/llvm/llvm-project/pull/156090

### DIFF
--- a/api/vscode-clangd.d.ts
+++ b/api/vscode-clangd.d.ts
@@ -25,6 +25,15 @@ export interface ASTParams {
   range: vscodelc.Range;
 }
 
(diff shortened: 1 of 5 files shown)


Febbe:
@HighCommander4 this draft can be used, to get a sneak peek into the search AST feature.

My plan is, to add a new view container in the sidepanel with a searchbar (as @torshepherd proposed) instead of the conditionally shown sub view in the editor view container with the requirement, to input the query via a floating dialog.

